PR #792, opened 2017-03-07
Add italian

LAfricain:
Adding italian In language choices.


karlkleinpaste:
i was going to get this into the just-released 4.0.5, but then i actually looked at this pull request for the first time. it is only the diff for the LINGUAS file; it does not include the it.po file.  if you will update to include it, i will be sure it goes into the next release.

LAfricain:
OK,
I translated (a little bit) on Launchpad, see the attached file, but I 
work mostly on the sword locale file, that means the books translation.

karlkleinpaste:
domcox handled this in april.

PR #808, opened 2017-07-13
avoid HAVE_GTK_, use GTK_CHECK_VERSION

yetist:
There are GTK_CHECK_VERSION macros in both GTK2 and gtk3, Can we use this instead of the HAVE_GTK_ definition?


karlkleinpaste:
i'll make a change to wscript as well, to stop the version stamp checks since we no longer need them.

Issue #810, opened 2017-07-16
Xiphos 4.05


jadesoye:
The commentary and dictionary panes are not showing, they are showing under the Bible text pane. Can someone fix this problem? Thanks
![screenshot from 2017-07-16 13-26-07](https://user-images.githubusercontent.com/30211719/28247547-8eecef08-6a2a-11e7-9c55-6c36af92b522.png)

yetist:
I think this problem may be related to WebKit2GTK(libwebkit2gtk-4.0). It seems that GtkPaned didn't get the width of this widget, it seems that its width has been 0. Maybe should check the code in the src/webkit directory.

jadesoye:
I'm using Fedora 26. There is no libwebkit2gtk-4.0 package in fedora 26. I installed webkitgtk3 but still the same problem. Any help to solve the problem is welcome. Thanks!

yetist:
Please test #811 

![2017-07-27 16-12-03](https://user-images.githubusercontent.com/665058/28661204-844234fc-72e8-11e7-934c-1e1920c1b7ea.png)

karlkleinpaste:
apologies for delay in handling, but it's fixed from 811 merge now.

Issue #756, opened 2016-04-25
WebKit2GTK+ Does not display properly under GTK 3.18 and GTK 3.20


cjbayliss:
I have been playing with this a little, and it seems that it is the GTK versions that cause the issue, and not a WebKit2GTK+ issue. I haven't tried under GTK 3.16

cjbayliss:
I spent a some time on this today, and can confirm that the problem is not WebKit2GTK, it seems to be related to how we pack the GTK widgets, maybe the XML GtkBuilder files are too depreciated for GTK 3.20?

cjbayliss:
![Screen Shot of the problem](http://i.imgur.com/dKIM81l.png)

cjbayliss:
So the problem seems to be related to the notebook tabs.

I can't work on this issue for about a month because I have used most of my internet.

greg-hellings:
That screenshot matches what I have in Fedora now. I was forced to switch to GTK3 builds for upcoming packaging changes, and your screenshot matches what I and other Fedora users are seeing.

creffett:
@greg-hellings did you find any solution for that? I'm trying to switch Gentoo to GTK3/webkit2 as well.

pacho2:
Please note Fedora already moved to webkitgtk4 for 4.0.5 to get rid of obsolete/vulnerable webkitgtk versions

karlkleinpaste:
fixed merged from 811.

PR #811, opened 2017-07-27
Fixed issue #810

yetist:
Please see issue #810

### DIFF
--- a/src/gnome2/main_window.c
+++ b/src/gnome2/main_window.c
@@ -1029,9 +1029,7 @@ void create_mainwindow(void)
 
 	// widgets.vpaned2
 	widgets.vpaned2 = UI_VPANE();
-	gtk_widget_show(widgets.vpaned2);
 	gtk_widget_set_size_request(widgets.vpaned2, 50, -1);
-	gtk_paned_pack2(GTK_PANED(widgets.hpaned), widgets.vpaned2, TRUE, FALSE);
 
 	// widgets.vbox_text
 	UI_VBOX(widgets.vbox_text, FALSE, 0);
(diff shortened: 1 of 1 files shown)


yetist:
Now it works.
```
$ ldd ./build/default/src/gnome2/xiphos|grep gtk
	libgtkhtml-editor-4.0.so.0 => /usr/lib/libgtkhtml-editor-4.0.so.0 (0x00007f5a69d3f000)
	libgtkhtml-4.0.so.0 => /usr/lib/libgtkhtml-4.0.so.0 (0x00007f5a69a68000)
	libwebkit2gtk-4.0.so.37 => /usr/lib/libwebkit2gtk-4.0.so.37 (0x00007f5a66e2e000)
	libgtk-3.so.0 => /usr/lib/libgtk-3.so.0 (0x00007f5a6652d000)
	libjavascriptcoregtk-4.0.so.18 => /usr/lib/libjavascriptcoregtk-4.0.so.18 (0x00007f5a63dd4000)

```

yetist:
This patch may have fixed issue #810 #806 #756. Has anyone tested it?

karlkleinpaste:
i will deal with all this soon.

karlkleinpaste:
finally looking at this, and regrets for delay.

but my initial reaction is to see that, with this patch, xiphos utterly ignores saved size preference. it always opens in default tiny state.

karlkleinpaste:
without the patch, that is, with code in latest master state, i rebuilt all of gtk2/wk1, gtk3/wk1, and gtk3/wk2 configurations. they all display normally.

somebody convince me that this problem was not a gtk3 or wk2 bug that has been fixed relatively recently.

yetist:
ok, now this patch can use the saved window size.

According to my test, this problem only occurs when gtk3 and WebKit2 are enabled.

I use archlinux, gtk3 version is 3.22.17, webkit2gtk version is 2.16.6.

karlkleinpaste:
fyi 4.0.6 to be released by the weekend.

yetist:
Good.
This patch just allows Xiphos to work on gtk3 and webkit2. But that may not be the best way.

According to the WebKit documents, The new WebKitWebView should not be displayed to the user until the “ready-to-show” signal is emitted. Please see [here](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#WebKitWebView-create).
And without this patch, the width and height of the WebKitWebView in my test is always 0.

PR #877, opened 2018-04-07
remove repeated arrow display

yetist:
Delete duplicate arrows on navbar_versekey and show that they exist only for visual effects.


karlkleinpaste:
thank you so much for your efforts. i will be dealing with the others shortly as well.

PR #883, opened 2018-04-15
fixed gtk_menu_popup_at_widget call

yetist:
Please review and test it.


OrdoFlammae:
To me, it seems like the checks that Travis CI initiated failed. It looks like most of the pull requests have failed the build system. Is there something that needs to be fixed in Travis? I'm not sure that it's good if all our pull requests (which seem to work fine) fail our checking system.

karlkleinpaste:
greg hellings informed me recently:

"CentOS is failing because I'm still waiting for the Sword 1.8 RPM to make it through the EPEL QA session...But for now you can tell that building on Fedora is still good."

PR #909, opened 2018-08-14
Touchup build process for ease across systems

alerque:
This is to make building from source and packaging  across alternate platforms easier. First it provides some feedback on the python2 requirement (easily missed on systems that default to python3) and uses a dynamic version string so that builds from HEAD actually have a meaningful version number.


alerque:
It looks like the Travis CI test for this failed, but looked at it and it seems like the _master_ branch is currently failing too. I'm happy to make any adjustments that need making to this so it can be accepted, but it's hard to check when the CI system isn't in working order.

greg-hellings:
There are conflicts between this and the master branch now, but overall this seems solid to me.

alerque:
@greg-hellings This _would_ have been solid if it had been accepted anywhere near the time I contributed it, but it is entirely obsolete now. It isn't just a conflict, the WAF based build has been dropped entirely and rebasing this on master and fixing conflicts literally ends up being a noop. I'm going to close it for now.

The concept of generating the version number at build time is still a good idea, but it will have to be re-implemented for the CMake based build system. I could probably do that easily enough for you in a separate PR, but I've had pretty rough luck getting _anything_ contributed to this project to date. Should I go for it?

Issue #894, opened 2018-04-30
code clean and optimize


yetist:
Maybe we should clean up and optimize  the code if have enough time? I found those code should be cleaned up:
1. #if 0 ...... #endif
2. The commented or invalided code.
3. Deprecations api migrate, like gconf to gsettings, dbus-glib to gdbus, etc.
4. Use gresource to include resource files, like images, ui files, etc.
5. Rewrite real object to instead alias or macro defined object under src/xiphos_html.
6. Eliminate global and external variables, decouple code, and avoid the internal implementation of exposed objects.
7. The address  of the license is located in the header of some files.

karlkleinpaste:
particularly for the "#if 0" case, that has at times been a habit of mine, to remove something from current use but keep it present (and visible, especially, so it's not forgotten) in case it needs to be brought back into use later. i recall using that a number of times some years ago, when doing a round of cleanups consequent to cppcheck analysis -- quite a few unused functions were kept within #if 0 blocks because i wasn't sure actual removal was a good idea. perhaps it's time to make them go away.

but otherwise, yes, it seems there is always a great deal of clean-up opportunity available. there is a lot of quite old code -- old elements of the settings structure, no longer used, for example. what was previously called gnomesword started in 2000, using gtk1, and dead artifacts from almost 2 decades ago remain.

teusbenschop:
About the deprecated 'dbus-glib', I am adding information from Debian that includes a recommendation and a porting guide. I hope it helps.


Source: xiphos
Severity: normal
Tags: sid bullseye
Control: block 895291 by -1
User: pkg-utopia-maintainers@lists.alioth.debian.org
Usertags: dbus-glib-deprecation

dbus-glib is a deprecated D-Bus library with some significant design
flaws, and is essentially unmaintained. I would like to minimize its
use, and eventually remove it from Debian. There will not be a
version that fixes its design flaws, because that would be a major
compatibility break, and any user of dbus-glib who is willing to port
their application to a newer, incompatible version should instead be
porting their application to a better D-Bus implementation such as
GDBus.

For most purposes, the recommended replacement for dbus-glib is the
GDBus family of APIs in GLib, found in <gio/gio.h>. This does not add
an additional dependency, because dbus-glib already depends on a
sufficiently new version of GLib. A porting guide is available in the
GLib documentation:
<https://developer.gnome.org/gio/stable/ch35.html>. Practical
examples of porting from dbus-glib to GDBus can be found in the git
[...]

PR #913, opened 2018-08-14
Remove all #if 0...#endif and some other dead code

alerque:
This should be a complete no-op code wise. Nothing in this commit should
change the function of the program in any way. If `git bisect` or
similar ever traces a problem to this commit, then something is wrong
and I didn't match start/end markers properly.

This partially addresses issue #894, bullet point 1 & part of 2.


alerque:
I just forced-pushed an amended commit to poke the CI builds which were not working upstream at the time I submitted this. Exactly 0 code was changed between commits, and this still merges cleanly to master.

alerque:
@greg-hellings Continuing conversation from #965 here (see my objection to the differences in the duplicate PR).

The [last expression I saw](https://github.com/crosswire/xiphos/issues/894#issuecomment-385499419) from @karlkleinpaste suggested that he might be warming up to this idea.

> perhaps it's time to make them go away.

src/main/display.cc
```diff
@@ -1184,28 +1181,7 @@ GTKChapDisp::getVerseBefore(SWModule &imodule)
...
-		buf = g_strdup_printf("%s<br />", cVerse.GetText());
-#else
-# if 0
-		buf = g_strdup_printf("%s<br />",
-				      (strongs_or_morph
-					   ? block_render(imodule.renderText().c_str())
-					   : imodule.renderText().c_str()));
-# else
-		// gosh this is gross.
+		// begin gosh this is gross.
```
yetist: Why you do this unrelated changes?

src/main/display.cc
```diff
@@ -1239,8 +1215,9 @@ GTKChapDisp::getVerseBefore(SWModule &imodule)
 		buf = g_strdup_printf("%s<br />",
 				      (strongs_or_morph ? block_render(divBuf->str) : divBuf->str));
 		g_string_free(divBuf, TRUE);
-# endif
-#endif /* !0 */
+
+    // end grossness
+
```
yetist: Why you do the unrelated changes?

src/main/display.cc
alerque: It isn't unrelated. Without the wrapping `#if 0 ... else ... end` block that got removed, it was no longer clear what section of code the remaining comment referred to. I defined the region to not loose the little bit of information provided by the fact that this code was previously grouped together.

src/main/display.cc
alerque: Not unrelated. See [this explanation](https://github.com/crosswire/xiphos/pull/913#discussion_r360250752).

alerque:
Ping. @greg-hellings or whoever, do we have any progress on this? Every time I open up the code to review something I bump into these dead blocks and get frustrated. It's really a lot easier on contributors if dead code is left in git history instead of in the files (and if PRs get reviewed and merged within at least a few years).

greg-hellings:
@alerque - I'm perfectly comfortable with this being merged, but when I had made the same proposal many years ago @karlkleinpaste  had expressed his desire to keep this code in place rather than in git history. I would rather not hit the Merge button over his wishes without hearing from him directly that it's OK. If he wants this merged, all he needs to do is click the "Merge pull request" button, so if he hasn't then I don't feel right going over his choice. He's the project lead.

karlkleinpaste:
i'll just mention that i'm about to merge this, but FYI i'm going to be doing a little more beyond this to get rid of the "gosh this is gross" section specifically because of more recent understanding about self-closing div and get rid of the gross hack entirely.

as for keeping around old #if 0 blocks... in my defense, it was a habit learned of much brutal experience literally decades in the past, and prior to the advances of things like git. it no longer matters as much to me as it once did.

alerque:
Thanks @karlkleinpaste. I cut my teeth writing code before versions control too. I have no idea how I survived those dreadful years.

PR #965, opened 2019-10-03
drop dead code

yetist:
Remove code in if 0.


greg-hellings:
I have suggested this long, long ago and @karlkleinpaste had pushed back against it, not wanting to do it. I don't know if his opinion has changed or not.

alerque:
This is also a duplicate of #913 in which I did the same work. This really needs to get merged at some point. The code is available in git history if somebody needs to resurrect something, but the amount of cruft in the source code is a barrier to contribution. It makes understanding what's what and finding problems much more difficult than need be.

greg-hellings:
I agree - but since he's expressed hesitation in the past, it falls to Karl to make that call

alerque:
I just reviewed this PR vs. my previous one. They are generally the same, but mine sticks much closer to the goal of just removing the `#if 0` code. This PR also changes some other things such as rewriting comments, adding and removing whitespace lines, and even removing code comments that explain what still active code is there for.

I strongly support removing all the `#if 0` blocks, but suggest that this PR be closed in favor of my previous submission. If this PR is continued those extra changes should be reverted, split into separate commits, or at least justified in some way. As it stands I do **not** think this PR should be merged as it removes good code comments and makes unnecessary changes elsewhere that will make `git bisect` and other debugging tools unnecessarily hard to use in the future.

alerque:
**Adendum:** I apologize, the difference isn't quite as significant as I first thought. The extra removed comments I thought were for live code were dead for other reasons and a miss-match for the code they were left with. I've tweaked my PR to reflect that change as well. Good catch @yetist.

My objection to the other unrelated changes (not strictly removal, but rewriting existing comments and changing white-space) still stands. ;-)

yetist:
@alerque 
> My objection to the other unrelated changes (not strictly removal, but rewriting existing comments and changing white-space) still stands. ;-)

Can you tell me what they are? Thanks.

yetist:
I don't care whether to merge #913 or #965, I just want to remove this dead code so that it is easier to contribute code.

karlkleinpaste:
#913 was merged instead.

PR #972, opened 2019-12-08
clear USE_GTK_3 macro

yetist:
Please review and test.

### DIFF
--- a/src/editor/webkit_editor.c
+++ b/src/editor/webkit_editor.c
@@ -802,7 +802,7 @@ static GtkWidget *editor_new(const gchar *title, EDITOR *e)
 	buttons_state.nochange = 1;
 
 	gbuilder_file = gui_general_user_file(
-#ifdef USE_GTK_3
+#if GTK_CHECK_VERSION(3, 0, 0)
 	    "gtk_webedit.ui"
 #else
 	    "gtk2_webedit.ui"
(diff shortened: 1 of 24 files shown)


alerque:
What is the advantage of checking for this repeatedly at compile time instead of once at configure time?

yetist:
There is no need to define USE_GTK3 in the build system.

1. One reason is that the build may be failed.
For example, when gtk-3.8 is installed in the system, the build system detects gtk+-3.0.pc, and then defines USE_GTK_3.
But in the code, if USE_GTK3, we call the newly api that added after gtk-3.14.

2. Another reason is that we can clean up the code.

For example, after gtk-2.18, we can use `gtk_widget_set_can_default`, so we can delete other lines: 
```
-#if GTK_CHECK_VERSION(2, 18, 0)
 	gtk_widget_set_can_default(button, TRUE);
-#elif defined(USE_GTK_3)
-	gtk_widget_set_can_default(button, 1);
-#else
-	GTK_WIDGET_SET_FLAGS(button, GTK_CAN_DEFAULT);
-#endif
```

3. Enhance strong maintainability

If you use USE_GTK_3,  can you quickly know that the following changes are safe?
```
-#ifdef USE_GTK_3
 			gtk_combo_box_text_append_text((GtkComboBoxText *)ss.entryUpper, book);
 			gtk_combo_box_text_append_text((GtkComboBoxText *)ss.entryLower, book);
-#else
-			gtk_combo_box_append_text(GTK_COMBO_BOX(ss.entryUpper), book);
-			gtk_combo_box_append_text(GTK_COMBO_BOX(ss.entryLower), book);
-#endif
```

karlkleinpaste:
simply because of the age of this -- yes i know i let it sit for an indecent time -- and that it eliminates gtk2 entirely (i noted this in email to xiphos-devel a year ago) which is req'd for windows build, closing this.

PR #973, opened 2019-12-10
Improve xiphos-html code and clear webkit directory

yetist:
Drop webkit directory.
Rewrite xiphos-html use gobject.


greg-hellings:
I don't know about dropping webkit or why we would do that.

I can comment that this patch is VERY unnecessarily noisy. For one, you arbitrarily rename "xiphos_html" to "xiphos-html", which introduces lots of very unnecessary changes to the code. Secondly, you change all the macros from "XIPHOS_HTML_*" to "xiphos_html_*", which again means that the code outside of the directory is changing in all sorts of places that it has no need to change.

The result is a very noisy PR that is difficult to comb through and find the actual meat of the changes.

alerque:
@yetist Including incremental commits that separate concerns would make things like this easier to review. Refactoring existing code (say to rename a variable) should be in a separate commit from something that changes functionality, drops a dependency or feature, or whatever.

yetist:
I want to say that the current code is difficult to maintain, the `webkit` directory and `xiphos_html` are completely duplicated and ugly. Please forgive me.

yetist:
For example, this is an alias, but the actual implementation is in the webkit directory.
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/xiphos_html/xiphos_html.h#L35-L47

The following functions are no longer used and should be removed.
```
#define XIPHOS_HTML_SHUTDOWN wk_html_shutdownGreg Hellings, 9 years ago: • Now builds cleanly with Gecko or Webkit
#define XIPHOS_HTML_INITIALIZE wk_html_initialize
```
Members inside the WkHtml object are visible to others and cause abuse:
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.h#L51-L65

Too many macro definitions and judgments. In fact, `webkitgtk-1.0`, `webkitgtk-3.0` and `webkitgtk-4.0` all implement `webkit2`, so it is safe to use webkit2 directly here.
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L24-L32

Cross-operations between objects increase the complexity of the program and cause difficulties in maintenance：
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L43-L47
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L79
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L102
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L104
https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L163-L165


And the last question is that I use a dash to separate the prefix from the type name: xiphos-html.h and xiphos-html.c to align with the conventions of most GNOME libraries.

refdoc:
Is the duplication not simply a result of the attempt to keep Xiphos render platform agnostic?Since I have started co attributing to Xiphos (sometime end of 2004) there have been three different render engines in use if one solely counts the names, not the versions - gtkhtml, xul and now WebKit. Over long stretches Xiphos required two engines' support in order to run on all platforms. Sent from my Huawei Mobile-------- Original Message --------Subject: Re: [crosswire/xiphos] Improve xiphos-html code and clear webkit directory (#973)From: Wu Xiaotian To: crosswire/xiphos CC: Subscribed For example, this is an alias, but the actual implementation is in the webkit directory.

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/xiphos_html/xiphos_html.h#L35-L47

The following functions are no longer used and should be removed.

#define XIPHOS_HTML_SHUTDOWN wk_html_shutdownGreg Hellings, 9 years ago: • Now builds cleanly with Gecko or Webkit

#define XIPHOS_HTML_INITIALIZE wk_html_initialize



Members inside the WkHtml object are visible to others and cause abuse:

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.h#L51-L65

Too many macro definitions and judgments. In fact, webkitgtk-1.0, webkitgtk-3.0 and webkitgtk-4.0 all implement webkit2, so it is safe to use webkit2 directly here.

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L24-L32

Cross-operations between objects increase the complexity of the program and cause difficulties in maintenance：

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L43-L47

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L79

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L102

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L104

https://github.com/crosswire/xiphos/blob/1c1826a3c256ff456f23bf77110d1cc6286e2a66/src/webkit/wk-html.c#L163-L165

And the last question is that I use a dash to separate the prefix from the type name: xiphos-html.h and xiphos-html.c to align with the conventions of most GNOME libraries.



—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or unsubscribe.

[

{

"@context": "http://schema.org",

"@type": "EmailMessage",

"potentialAction": {

"@type": "ViewAction",

"target": "https://github.com/crosswire/xiphos/pull/973?email_source=notifications\u0026email_token=AADY6C5HMJHWB44CBNRWM6LQZR2LNA5CNFSM4JYZJ5X2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEHMF7MI#issuecomment-567828401",

"url": "https://github.com/crosswire/xiphos/pull/973?email_source=notifications\u0026email_token=AADY6C5HMJHWB44CBNRWM6LQZR2LNA5CNFSM4JYZJ5X2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEHMF7MI#issuecomment-567828401",

"name": "View Pull Request"

},

"description": "View this Pull Request on GitHub",

"publisher": {

"@type": "Organization",

"name": "GitHub",

"url": "https://github.com"

}

}

]

yetist:
Can you tell me if I deleted the code in the `editor` directory (lost html edit function), can I use `webkit2gtk` to run xiphos and read the bible on Linux, Windows and MacOS?

greg-hellings:
Personally I would prefer this be broken up into multiple PRs:

Renaming xiphos_html to xiphos-html should be its own change. If we're going to consider doing that, then we should consider that by itself.

If we're going to drop the webkit directory instead of having the parallel implementations, then that should be its own PR as well.

Currently there is too much going on here in one PR for me to fully grasp and comment on each change individually.

karlkleinpaste:
i'm going to roll with the critiques on this one. rather too much going on, and rather too much change that's not integral to the purpose. up- or down-casing for no identifiable reason doesn't feel appropriate, and naming macros in all uppercase (as is done for GTK macros themselves) fit better.

PR #975, opened 2019-12-11
Built-in small zip library instead of using libgsf

yetist:
Please review and test, thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ add_subdirectory(help)
 
 # create packages
 add_subdirectory(cpack)
+add_subdirectory(subprojects/zip)
 
 # report
 include (XiphosReport)
--- a/cmake/XiphosDependencies.cmake
+++ b/cmake/XiphosDependencies.cmake
@@ -87,7 +87,6 @@ pkg_check_modules(Gnome REQUIRED IMPORTED_TARGET
   "libsoup-2.4"
   "pango"
   "gconf-2.0"
-  "libgsf-1>=1.14"
   "libxml-2.0>=2.7.8"
   )
 
--- a/src/gtk/CMakeLists.txt
+++ b/src/gtk/CMakeLists.txt
@@ -211,6 +211,7 @@ target_link_libraries(xiphos
   Xiphos::Editor
   Xiphos::Webkit
   Xiphos::Xiphos_html
+  Subproject::zip
   )
 
 
--- a/src/gtk/utilities.c
+++ b/src/gtk/utilities.c
@@ -45,6 +45,7 @@
 #include "gui/mod_mgr.h"
 #include "gui/widgets.h"
 #include "gui/dialog.h"
+#include "zip.h"
 
 #include "main/lists.h"
 #include "main/mod_mgr.h"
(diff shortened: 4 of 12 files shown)


greg-hellings:
1) Why? What's wrong with libgsf?

2) Please don't include bundled libraries. It makes it an aboslute NIGHTMARE to maintain packages in distros (all of the ones I know of have a very strong prohibition against bundled libraries bceause of the risk of them getting out of date with bugs and vulnerabilities). As a downstream maintainer (Fedora/EPEL) I immediately have to take this change and undo it before packaging for Fedora and CentOS - at minimum I just include a reverse version of this patch, or I have to modify this code to build without the bundled code and instead link against system zlib instead. They're both a maintenance nightmare.

alerque:
Second what @greg-hellings said above about packaging, any time you bundle the source of another library into a project you add another ring to purgatory for distro maintainers. Using an external call to a major/well known package is by far preferred because it makes the overall system easier to secure.

Libgsf does not seem to be a problem that needs fixing. At least on distros I maintain packages for it's available in the core package set so it's an easy dependency. Grep spake for Fedora and CentOS. I can speak for Arch Linux, and off the top of my head I know Debian is the same way: we will have to patch this out of the source in order to package properly!

yetist:
What you said makes sense, we shouldn't embed third-party code. I meant to make a subproject, like [this](https://gitlab.gnome.org/GNOME/gtk/tree/master/subprojects), but I don't know how cmake does it.

About libgsf, when using the latest `freedesktop platform/sdk` 19.08 to make xiphos flatpak, libgsf compiles error. I Based on this:

https://github.com/flathub/org.xiphos.Xiphos/blob/74574044f42cb8c15cd04536f2d84d283cffe3a6/org.xiphos.Xiphos.json#L22-L31

But it seems that the latest version of libgsf still has problems.

Using libgsf in xiphos is only to generate zip files. Why can't we use smaller libraries?

greg-hellings:
I don't think there's an objection to using a smaller or different library. Just to embedding the target library into the app's code tree.

If generating a ZIP archive is the only reason we're using libgsf, then it's entirely reasonable to use system zlib or something else specific to that purpose. It should make building the flatpak easier, but should also make building packages everywhere faster and simpler, including cross-compiling for Windows.

alerque:
I don't have an attachment to libgsf either, or anything against using a "smaller" library. What I would be careful about is whether it is more or less "common". It is much better to use a big bulky library to do a simple job if that library is more likely to already be packaged for or even already installed on systems. If there is something lighter weight / faster / more ubiquitous it would be great to use that, just not at the expense of running into more places without up to date packaging.

And yes, either way don't bundle the source in this project's source tree. Platforms than require bundled dependencies can handle that other ways.

refdoc:
Isn't there zip code in libsword?

greg-hellings:
Our CI system is back up and running. If you want to re-submit this as a change to depend on system zlib instead of bundle a zip library in Xiphos, I'd be happy to press it forward or bother @karlkleinpaste about reviewing this.

karlkleinpaste:
this is one of those "i appreciate the intent, and the effort" things, but given the discussion about ill-advised inclusions into the tree, i'm going to decline this one. it's rare that a proposed change is immediately critiqued the way this one was. the idea's fabulous, if you think libgsf is too big a weapon to wield for its single purpose in xiphos. a different solution would be fine, just not as offered.

PR #979, opened 2019-12-20
Generate version string at build time from git tags

alerque:
This is a replacement PR for #909 which was obsoleted by the switch from WAF to CMake.

The VERSION string will be automatically generated and updated based on the git tag status.

- [x] Local builds from git clones
- [x] CI builds
- [x] Source tarballs that don't include git repository information

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,13 @@ sudo: required
 language: bash
 services:
   - docker
+git:
+  depth: false
 
 before_install:
   - curl -Ls -o docker-build https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/docker-build
   - chmod +x docker-build
+  - git fetch --tags
 
 install:
   - sudo apt-get install -y python3-pip python3-setuptools
@@ -32,6 +35,7 @@ requires:
     - appstream-glib
     - autoconf
     - automake
+    - git
     - gconf
     - cmake
     - dbus-glib
@@ -62,6 +66,7 @@ requires:
     - cpp
     - desktop-file-utils
     - fp-utils
+    - git
     - gsettings-desktop-schemas-dev
     - intltool
     - itstool
@@ -88,6 +93,7 @@ requires:
     - cmake
     - dbus-glib-devel
     - desktop-file-utils
+    - git
     - gcc-c++
     - gconfmm26-devel
     - gtk3-devel
@@ -110,6 +116,7 @@ requires:
     - cpp
     - desktop-file-utils
     - fp-utils
+    - git
     - gsettings-desktop-schemas-dev
     - intltool
     - itstool
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,10 +57,17 @@ endif()
 list (APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 list (APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
-# parse the current version based on git tags
-include(GetGitRevisionDescription)
-git_describe(GIT_VERSION --tags)
-string(REGEX REPLACE "-([0-9]*).*" ".\\1" VERSION "${GIT_VERSION}")
+# parse the current version based on git tags, then cache a result on behalf
+# of source tarballs that don't have the repo .git metadata
+set(SOURCE_VERSION_CACHE "${PROJECT_SOURCE_DIR}/cmake/source_version.txt")
+if(EXISTS "${PROJECT_SOURCE_DIR}/.git")
+	include(GetGitRevisionDescription)
+	git_describe(GIT_VERSION --tags)
+	string(REGEX REPLACE "-([0-9]*).*" ".\\1" VERSION "${GIT_VERSION}")
+	file(WRITE ${SOURCE_VERSION_CACHE} "${VERSION}")
+else()
+	file(STRINGS ${SOURCE_VERSION_CACHE} VERSION)
+endif()
 
 # configure project name and version
 project (xiphos
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,12 +21,8 @@
 cmake_minimum_required (VERSION 3.6 FATAL_ERROR)
 cmake_policy(VERSION 3.6)
 
-
-### set project name and version HERE ###
-project (xiphos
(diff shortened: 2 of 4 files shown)


alerque:
This working now for 2/3 cases now. I'm not sure where to put the information for the latter case, but I'm pretty sure we just need to ask CMake to place it somewhere for bundling, and then use it from that place in the event git meta information is unavailable.

If somebody lets me know what command(s) you run to make a source release tarball I'll figure out where that can/should be.

.travis.yml
```diff
@@ -4,12 +4,15 @@ sudo: required
...
   - docker
+git:
+  depth: false
 
 before_install:
   - curl -Ls -o docker-build https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/docker-build
   - chmod +x docker-build
 
 install:
+  - git fetch --tags
```
greg-hellings: To my mind this is more of a `before_install` command.

alerque: Good call, I was referencing a project that didn't have a `before_install` section at all, but that does make more sense in this case.

domcorbex:
> If somebody lets me know what command(s) you run to make a source release tarball I'll figure out where that can/should be.

Hi!

About source tarballs, 
(From [Install.md](https://github.com/crosswire/xiphos/blob/master/INSTALL.md))
The following command will generate `zip` and `tar.gz` source files:
    `$ make package_source`

For the code, see [Cpack/CMakeLists.txt](https://github.com/crosswire/xiphos/blob/master/cpack/CMakeLists.txt)

greg-hellings:
CI is significantly updated, now. If you wanted to close this off and push updates, then we can have them tested against current CI

alerque:
I'm actually got back to this and started trying to figure out the last piece. See [this SO question](https://stackoverflow.com/questions/60726855/how-do-i-hard-code-a-version-number-into-a-cmake-cpack-with-dynamic-version-number). When I figure out the right way to do this I'll rebase or merge or something to get CI checks up to date here. Thanks.

CMakeLists.txt
```diff
@@ -61,6 +57,15 @@ endif()
...
 list (APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
+# parse the current version based on git tags
+include(GetGitRevisionDescription)
+git_describe(GIT_VERSION --tags)
+string(REGEX REPLACE "-([0-9]*).*" ".\\1" VERSION "${GIT_VERSION}")
+
+# configure project name and version
+project (xiphos
+	VERSION "${VERSION}")
```
greg-hellings: If this isn't working then maybe try the set_property command?

alerque: I guess it depends of what you mean by "working". It's setting the version properly: the built software gets the right version, the logs show show the right version, and the packaged filenames have the right version. The problem is that it then bundles a copy of itself inside the source package, and _THAT_ can't figure anything out because `VERSION` isn't actually set anywhere it is derived from git repository meta data — which of course doesn't exist in the tarball.

From my experience with GNU-Make the way to fix this is probably to cache the version in a file somewhere during normal builds and use the cache to retrieve it in the event `.git`, I just don't know how to do that in Cmake.

greg-hellings: Ah, yes. It would be pretty normal to set the value statically when you're cutting a release. I would suggest writing a file into the cmake/ directory. Then you can check for its existence and use that if found. If you don't find it, then use the git method as a fallback.

alerque: I think I got it. Finally.

karlkleinpaste:
@greg-hellings have you taken a brief look at this? i'm inclined to merge but i'm not even going to pretend that i understand many of the vagaries of cmake used. from my very brief perusal, do i correctly understand that future version numbers are to be placed in a file cmake/source_version.txt?

greg-hellings:
@karlkleinpaste I've looked at it. The way I understand its logic (I'm just assuming the large included files do what they say on the tin, I haven't deeply hand-traced through them) in approximate shell-and-pseudo-code terms:
```
if [ -d .git ]; then
  version=$(get_version_from_git)
  echo ${version} > cmake/source_version.txt
else
  version=$(cat cmake/source_version.txt)
fi
set_version $version
```

So when cutting a release we'll just need to echo the version number into cmake/source_version.txt where it will get picked up and set accordingly for the rest of the build process. Development versions will be based on the latest tag. No more needing to manually set the value in CMakeLists.txt and update it after each release.

It looks good to me.

alerque:
> So when cutting a release we'll just need to echo the version number into...

Actually this is incorrect. The whole point of this is to _NOT_ have the version hard coded in the source in any file at all. You don't have to do anything at all, the version is auto generated at build time based on the sate of the git repository.

To gut a new release you first `git tag X.Y.Z`, then do a clean build. The build should have the right version baked in, and doing a source build will hard code the version in where it needs to be so that you can build from that source without it being a git repository.

Development builds between tags will have versions relative to the last tag.

greg-hellings:
Hmmm, but building is not typically part of a release process. With git you're most frequently going to spit out the tarball with "git archive --prefix=xiphos-4.2.0/ -o ../xiphos-4.2.0.tar.gz master" or a similar command.

alerque:
> Hmmm, but building is not typically part of a release process.

It should be, source distributions should be prepared by the build system. Cmake has this functionality built in. After running `cmake -S . -B build -DCMAKE_BUILD_TYPE=Release ...` then you generate source tarballs using `make -C build package_source`. This is pretty fundamentally different that running from a tarball of a git working tree. You can _either_ build from a git clone _or_ from a source package, but there are a number of reasons (besides this even) why you should be posting source tarballs built with the proper tooling.

Note CMake works this way, GNU Automake is setup to work this way, Python setup_tools is setup to work this way, and more.

If you'd like as a follow up when this is merged I can help setup Github Actions to do this automatically and post sources when releases are tagged.

greg-hellings:
Ah, I was unaware of `make package_source` from CMake. That should make cutting releases trivial.

I've been having great luck with Github Actions in my $DAY_JOB doing everything from linting, CI, and even CD of things like Python packages and Ansible collections. We can absolutely put together something that generates all the different package types automatically from a tagged release.

alerque:
@greg-hellings I just setup another project to post build artifacts to tagged releases via Actions yesterday, after this is merged I'll have a look at doing the same for Xiphos.

PR #986, opened 2020-03-15
remove gconf-2.0 from xiphos (after CMake)

### LINKED ISSUE #918: Debian will drop gconf2 - patch attached
There was a bug report in Debian.
It is on the Xiphos package.

Here is the text of the bug report:

Your package depends and build-depends on gconf, but gconf will be
removed from Debian soon (most probably right after the Buster
release).

Furthermore, it seems that the dependency is completely unnecessary --
xiphos installs URL handlers but these keys are never read by GNOME as
this functionality has been deprecated long time ago.  The right way
to achieve this is through .desktop/mime/aplication-registry files;
that would make them work under any XDG-compliant environment.

Here is the link to the report:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=912807

The bug report contains a partial patch.

mcepl:
We should finally push this issue to some resolution, so that all Linux distros are not carrying this patch.

Fixes #918

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,6 @@ requires:
     - appstream-glib
     - autoconf
     - automake
-    - gconf
     - cmake
     - dbus-glib
     - docbook-utils
@@ -66,7 +65,6 @@ requires:
     - intltool
     - itstool
     - libdbus-glib-1-dev
-    - libgconf2-dev
     - libglade2-dev
     - libgsf-1-dev
     - libenchant-dev
@@ -89,7 +87,6 @@ requires:
     - dbus-glib-devel
     - desktop-file-utils
     - gcc-c++
-    - gconfmm26-devel
     - gtk3-devel
     - gtkhtml3-devel
     - intltool
@@ -116,7 +113,6 @@ requires:
     - libdbus-glib-1-dev
     - libenchant-dev
     - libgail-3-dev
-    - libgconfmm-2.6-dev
     - libglade2-dev
     - libgsf-1-dev
     - libgtk-3-dev
--- a/cmake/XiphosDependencies.cmake
+++ b/cmake/XiphosDependencies.cmake
@@ -86,7 +86,6 @@ pkg_check_modules(Gnome REQUIRED IMPORTED_TARGET
   "gobject-2.0"
   "libsoup-2.4"
   "pango"
-  "gconf-2.0"
   "libgsf-1>=1.14"
   "libxml-2.0>=2.7.8"
   )
--- a/src/gtk/gui.c
+++ b/src/gtk/gui.c
@@ -68,9 +68,6 @@ void gui_init(int argc, char *argv[])
 	if (!gtk_init_with_args(&argc, &argv, NULL, NULL, NULL, NULL)) {
 		exit(1);
 	};
-#ifndef WIN32
-	gconf_setup();
-#endif
 #ifdef HAVE_DBUS
 	ipc = ipc_init_dbus_connection(ipc);
 #endif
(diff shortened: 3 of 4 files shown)


greg-hellings:
If memory serves (please advise, @karlkleinpaste) the main reason GTK2/GConf2 is kept around is that Windows builds are still done from that stack.

alerque:
@mcepl Thanks for preparing this, I was just sitting down to do the same.

@greg-hellings If that turns out to be the case there should be some platform detection in the cmake file and Linux platforms shouldn't bump into build or run errors when it doesn't exist.

mcepl:
I don’t think those failed builds are my problem: something with Biblesync, isn’t it?

alerque:
@mcepl At least on Arch, you are neither installing the packages (not in [community] but available to build in AUR or prebuilt in my repository) nor installing it from source. The source build is limited to Ubuntu. You need to do one of those at least.

As for Ubuntu ... is it possible to just fetch and install the Debian package instead of building from source? The problem with the source build is likely [this two year old bug](https://github.com/karlkleinpaste/biblesync/issues/9) that has been fixed in the repository but never released.

alerque:
@mcepl I'll submit a fix for the Arch build as a PR to your branch here.

alerque:
Hmmm, some of that is not an issue with the PR so much as recent changes in master. I'll submit a PR there, then this can be rebased.

mcepl:
> Hmmm, some of that is not an issue with the PR so much as recent changes in master. I'll submit a PR there, then this can be rebased.

That’s what I meant. Thank you.

alerque:
@mcepl I fixed the Arch Linux side of things is #987. When that is merged this PR can be rebased. Since I was merging to master I had to go ahead and build/install gconf. If you rebase this (or anytime after this is merged) there is one line in the Travis config to drop that will remove the gconf build & install.

greg-hellings:
I merged this, after fixing conflicts with existing CI configurations.

karlkleinpaste:
an after-the-fact observation.

as i return from the dead, as it were, and i see this is already done, the question on my mind that i will need to pursue in the next couple of days is how i'm going to build for windows, given that gtk3 from mingw packages made for a truly terrible display in the past. the gtk configuration (gtkrc and so forth) worked well for gtk2, but the similar machinery for gtk3 was never there.

has anyone but me built for windows after this change? has anyone but me built for windows in the last 5 years?

greg-hellings:
> has anyone but me built for windows after this change? has anyone but me built for windows in the last 5 years?

After this change? no
In the last 5 years? yes

I don't know anything about styling GUIs. I'm not exactly picky about my displays so I don't bother with customizing them. It would be interesting to see how a native GTK3 build looks, these days.

domcorbex:
> has anyone but me built for windows in the last 5 years?

I checked/fixed the build for windows with CMake until Fedora 30, but I didn't on F31.

domcorbex:
I failed to build Xiphos for windows on Fedora 31.
Unfortunately webkitgtk has been orphaned 6 weeks ago:
https://src.fedoraproject.org/rpms/mingw-webkitgtk/c/4e52fea898a52cf7dc55b635c447d958cd2d7510?branch=master
so mingw-webkitgtk packages are gone as well.

Moreover, I could not find a mingw-webkit2gtk package, so I can't even try to build a GTK3 version of Xiphos for windows.

greg-hellings:
@domcorbex - can you drop your directions in here for what you tried. I'm going to see about updating the webkitgtk MinGW package in Fedora, if I can get it to build. At least for us, privately, to build Xiphos that is.

domcorbex:
@greg-hellings - This is how I proceed: 
1. Create and enter a toolbox
2. Install build tools
3. Install mingw[32|64] dependencies
4. Build and install Biblesync with Mingw
5. Configure and build Xiphos
6. Exit the toolbox and test the .EXE in a vm.

Detailed instructions are in INSTALL.md, in the _Cross-Compiling Xiphos for Windows_ section that I wrote myself.

The missing package is mingw[32|64]-webkitgtk.noarch:
`Error: Unable to find a match: mingw32-webkitgtk.noarch`

So yes, we can try to build ourselves mingw[32|64]-webkitgtk.noarch packages, but it is an approach that leads to a dead end, webkitgtk is insecure and no longer maintained upstream.

For future versions of Xiphos for Windows, we need an up-to-date mingw-webkit2gtk package. That implies a build against GTK3 (3.22 actually), something I could never achieve.

Hopefully, the GTK website has been redesigned in february and hosts now install instructions for GTK on Windows at https://www.gtk.org/docs/installations/windows/

PR #987, opened 2020-03-17
Fix Arch Linux Docker setup

alerque:
* Use Official Docker image
* Fix required package list
* Re-include in per-distro Biblesync source install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
   - ./docker-build --name ${DISTRO} -vvvv --config .travis.yml --build scripts
 
 env:
-  - DISTRO="archlinux/base"
+  - DISTRO="archlinux:latest"
   - DISTRO="debian:sid"
   - DISTRO="fedora:latest"
   - DISTRO="ubuntu:16.04"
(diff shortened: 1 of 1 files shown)


alerque:
@greg-hellings This fixes the completely broken Arch Linux end of things in Travis. I didn't attempt Ubuntu yet, nor did I try to cleanup other odds and ends like the MAKEFLAGS I commented on earlier, it just fixes the Docker image for Arch so that all the deps are in order, and the build now passes.

alerque:
@greg-hellings Just for quick reference the issue with all the Ubuntu builds [seems to be](https://travis-ci.org/github/crosswire/xiphos/jobs/663450551#L1366) their decrepit `cmake` version:

    CMake 3.7.0 or higher is required.  You are running version 3.5.1

I'm _sure_ somebody has run into this and has a working source install or PPA with newer cmake, **but**...

Given that users will also bump into this problem I'm wondering what in the cmake setup actually requires >= 3.7. Any idea?

alerque:
Never mind that last question. It's not even Xiphos's build that is failing, it is SWORD's. Why they need newer `cmake` I don't know, or why Xiphos needs 3.6 I don't know, but it doesn't really matter. We have to put a newer `cmake` on the system to get this testing done on Ubuntu.

greg-hellings:
I have Fedora and Debian working locally.

Xiphos' CMake system actually requires CMake 3.11. The latest LTS version of Ubuntu has 3.10. So it looks like we might end up not supporting Ubuntu LTS older than 20.04 in the next release. In my opinion, that's OK.

alerque:
> Xiphos' CMake system actually requires CMake 3.11.

In that case I think the `cmake_policy()` setting needs updating. For my PR #979 I'll need at least 3.6, but if something needs newer than that policy should be updated to reflect as much.

alerque:
Nevermind, I see you _did_ update policy. I was looking at the wrong branch.

PR #992, opened 2020-03-27
Locale

LAfricain:


LAfricain:
Il closed an other pull request #969 bucause this one is more complet.

alerque:
In the future note you can extend or teak the contents of Pull Requests while they are open just by pushing more commits to the same branch.

LAfricain:
Thank you for the advice. Yes it's better to do like you proposed.

LAfricain:
I wanted to change also cop, but a commentary says issues happen with the font. Is this still relevant?

ui/languages
```diff
@@ -2469,8 +2482,8 @@ gab	Gabri
 gad	Gaddang
 gae	Guarequena
 gaf	Gende
-gag	Gagauz
-#gag-Cyrl	Gagauz
+#Gagauz
+gag	Гагаузы
```
alerque, line 2486: This entry is incorrect, the reason there is a `-Cyrl` sub-locale in that the default Gagauz alphabet is a Latin based one similar to Turkish.

ui/languages
```diff
@@ -5537,6 +5561,8 @@ oru	Ormuri
 orv	Old Russian
 orw	Oro Win
 orx	Oro
+# Orya
```
alerque, line 5564: Why have Latin comments for a handful of these entries an not others? Can we just drop the English comments, this is a locale list, the codes should identify them.

ui/languages
```diff
@@ -6151,9 +6177,9 @@ sad	Sandawe
 sae	Sabanês
 saf	Safaliba
 sag	Sango
-sah	Yakut
-# as above, in cyrillic
-sah-Cyrl	Yakut
+# Yakut
+sah	Саха тыла
```
alerque, line 6181: This entry should not be Cyrillic, that's why there is a `-Cyrl` variant, no?

ui/languages
LAfricain: I have no opinion about this. The old entries had comments with the english name, but not the 3 letters entries. I added the English comments when it was in cyrillic.

PR #1004, opened 2020-04-17
Fix build errors for minizip

greg-hellings:
Fixes build errors related to minizip - doesn't seem to write the files properly, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,6 @@ requires:
     - gtk3
     - gtkmm
     - intltool
-    - libgsf
     - libxml2
     - minizip
     - python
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -46,6 +46,7 @@ requires:
     - intltool
     - libgsf
```
alerque, line 47: Besides adding minizip, libgsf should be removed from the dependencies for all these builds just to check that it works without it.

.travis.yml
```diff
@@ -141,6 +144,7 @@ requires:
     - libglade2-dev
     - libgsf-1-dev
```
alerque, line 145:
```suggestion
```

greg-hellings: Good call. I've removed all the gsf's from the build deps. New test up.

alerque:
Ooops, sorry. I had those code reviews written earlier but they didn't post for some reason. I guess we can disregard ... except that zlib is now also a dependency for all platforms. I'm not too worried given in is pretty ubiquitous.

PR #1007, opened 2020-04-18
use GTK_CHECK_VERSION instead of USE_GTK_3

yetist:
Let's test the Travis CI.


yetist:
CI not work.

alerque:
CI is running, for example [this PR's results are here](https://travis-ci.org/github/crosswire/xiphos/builds/676618439). For some reason it is not reporting back to Github. I suspect this is because the repo permissions have not been granted to the latest version of the app. Somebody with owner privs on this repo needs to re-authorize Travis using the .com domain app not the old .org one.

PR #1012, opened 2020-04-25
Win32 builds

greg-hellings:
Converts win32 build script to run directly on any host/container, including directions on how to run it.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,0 +1,41 @@
+name: builds
+
+on:
+  push: {}
+  pull_request: {}
+
+jobs:
+  build_windows:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - "-win32"
+          - "-win64"
+    steps:
+      - name: checkout Xiphos
+        uses: actions/checkout@v2
+        with:
+          path: xiphos
+          fetch-depth: 0
+      - name: fetch history
+        run: |
+          cd xiphos && git fetch origin +refs/tags/*:refs/tags/*
+      - name: checkout Biblesync
+        uses: actions/checkout@v2
+        with:
+          repository: karlkleinpaste/biblesync.git
+          path: biblesync
+      - name: build windows targets
+        shell: bash
+        id: build
+        run: |
+          set -ex -o pipefail
+          docker run -i --rm -v "$(pwd):/source" fedora:30 /source/xiphos/win32/xc-xiphos-win.sh ${{ matrix.version }} -b=/source/biblesync
+          echo "::set-output name=artifact::$(ls xiphos/*.exe)"
+      - name: upload artifact
+        uses: actions/upload-artifact@master
+        with:
+          name: packages
+          path: ${{ steps.build.outputs.artifact }}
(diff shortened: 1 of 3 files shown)


greg-hellings:
If you check the branch in my fork you'll see that these builds are passing in Github Actions and producing artifacts (small link in the upper right part of the Actions page) which can be downloaded and tested on a Windows machine. I completely lack access to Windows at the moment, so I can't do that testing.

karlkleinpaste:
looks good to me but let's wait on dom to see it as well.

karlkleinpaste:
> builds are passing in Github Actions and producing artifacts (small link in the upper right part of the Actions page)
i'm looking at your actions page and have no idea what you're talking about.
screenshot1

greg-hellings:
That list in the center is all the times the task has run. If you click on the title of that latest run "Suppress setting WIN32 at all times" then there will be a link to artifacts titled "packages".

Turns out it's in the dead center of that page. It's only in the corner if you're looking at the logs of individual build steps.

artifacts

karlkleinpaste:
got it.

detail 1: it says "59M" on the packages.zip download. it actually was 120. am i to assume that this is because there are 2 installers present?

detail 2: with wine, i ran the 64 installer and the result runs and looks all right (as far as "all right" goes under wine), except for one thing, the sword against which it's built is not up to SVN 3721, so the recent fix for pre-book and -chapter material is taking over the entire chapter in ESV. any possibility we can get a patch release of sword for mingw? otherwise, you need to revert
https://github.com/crosswire/xiphos/commit/74c88b94dc840d2e72d6a4ac974c0b3c56c42647.diff
in order to keep it sane.

detail 3: it's built to show a pseudo-console. for a release build, it needs whatever that option is to turn that off.

if you've got wine installed, you can try it, too.

good work, both of you. thanx.

domcorbex:
> detail 3: it's built to show a pseudo-console. for a release build, it needs whatever that option is to turn that off.

Option to add: 
`-DCONSOLE=OFF`
Default is ON

greg-hellings:
> got it.
> 
> detail 1: it says "59M" on the packages.zip download. it actually was 120. am i to assume that this is because there are 2 installers present?
> 

I assume so. Since they're built in parallel and are each uploaded from different threads to the same artifact group, I imagine that it's calculating size based on the first one that uploads and doens't update its metadata after the second file gets added. I plan to move the entire Travis CI runs over to this same file once we're done with this PR, so that will eventually grow to include the RPM file and probably source tarball as well.

> detail 2: with wine, i ran the 64 installer and the result runs and looks all right (as far as "all right" goes under wine), except for one thing, the sword against which it's built is not up to SVN 3721, so the recent fix for pre-book and -chapter material is taking over the entire chapter in ESV. any possibility we can get a patch release of sword for mingw? otherwise, you need to revert
> [`74c88b9`.diff](https://github.com/crosswire/xiphos/commit/74c88b94dc840d2e72d6a4ac974c0b3c56c42647.diff)
> in order to keep it sane.

I'm using the F30 released package. I should be able to pull that commit and apply it. But in about 5 or 6 weeks I'm going to lose the power to add or update Fedora 30 packages. So, at that point, we'll have to eventually move to a source build of Sword like we're doing with Biblesync now, unless we figure out the toolkit situation and update to being able to run in latest Fedora.

As an addendum, I didn't realize mingw-biblesync was missing from Fedora. I've made a quick package review request for that, hopefully we can get it in before the F30 window closes. But I'm waiting on a review for that.


> 
> detail 3: it's built to show a pseudo-console. for a release build, it needs whatever that option is to turn that off.
> 

I just disabled it whole cloth for now. Figuring out how to detect release vs nightly builds will be something we can tackle down the road if we want console re-enabled for debugging purposes. In the meantime, it takes just a couple minutes to run a build locally now, so it should be possible to make a local debug build if it becomes imminently necessary.


> if you've got wine installed, you can try it, too.
> 
> good work, both of you. thanx.

PR #1015, opened 2020-04-28
More ci

greg-hellings:
CI for Linux added to branch, moved from Travis to Github Actions
Generate source and binary packages
Add continuous deployments whenever a tag is pushed to Github.

### DIFF
--- a/.ci-build.yml
+++ b/.ci-build.yml
@@ -133,7 +133,7 @@ variables:
 before_scripts:
   - |
     START_DIR="$(pwd)"
-    WORK_DIR="/tmp"
+    WORK_DIR="${TEMP}"
     if [ "${DISTRO_NAME}" == "archlinux" ];then
         cd "${WORK_DIR}"
         id -u aur_install &>/dev/null || useradd -r -m -U -G wheel -k /dev/null aur_install
--- a/.ci-build.yml
+++ b/.ci-build.yml
@@ -15,6 +15,7 @@ requires:
     - python
     - sword
     - webkit2gtk
+    - zip
     # base-devel group (for AUR installs):
     - autoconf
     - automake
@@ -40,7 +41,6 @@ requires:
     - sudo
     - texinfo
     - which
-    - zip
 
   debian:
     - appstream-util
@@ -91,6 +91,7 @@ requires:
     - libxml2-devel
     - make
     - minizip-devel
+    - rpm-build
     - sword-devel
     - webkitgtk4-devel
     - yelp
@@ -192,9 +193,13 @@ build_scripts:
     cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DGTKHTML=ON -DCMAKE_MAKE_PROGRAM=/usr/bin/make .
     make -C build -j$(nproc)
     make -C build mhelp package_source
-    echo "::set-output name=artifact::$(ls build/*.tar.gz)"
-    #if [ "${DISTRO_NAME}" == "fedora" ]; then
-    #    echo "::set-output name=artifact::$(ls *.rpm)"
-    #elif [ "${DISTRO_NAME}" == "ubuntu" ]; then
-    #    echo "::set-output name=artifact::$(ls *.deb)"
-    #fi
+    if [ "${DISTRO_NAME}" == "fedora" ]; then
+        (cd build && cpack -G RPM)
+        echo "::set-output name=artifact::$(ls build/*.rpm)"
+    elif [ "${DISTRO_NAME}" == "ubuntu" ]; then
+        (cd build && cpack -G DEB)
+        echo "::set-output name=artifact::$(ls build/*.deb)"
+    elif [ "${DISTRO_NAME}" == "archlinux" ]; then
+        (cd build && cpack -G TGZ)
+        echo "::set-output name=artifact::$(ls build/*.tar.gz)"
+    fi
--- a/.ci-build.yml
+++ b/.ci-build.yml
@@ -1,40 +1,3 @@
-# vim: set ts=2 sts=2 sw=2 expandtab :
-dist: bionic
-language: bash
-os: linux
-services:
-  - docker
-git:
-  depth: false
-
-addons:
-  apt:
-    packages:
-      - python3-github
-      - python3-pip
-      - python3-setuptools
-
-before_install:
-  - curl -Ls -o docker-build https://github.com/mate-desktop/mate-dev-scripts/raw/master/travis/docker-build
-  - chmod +x docker-build
-  - git fetch --tags
-
-install:
-  - sudo pip3 install --upgrade pip
-  - ./docker-build --name ${DISTRO} --config .travis.yml --install
-
-script:
-  - travis_retry ./docker-build --name ${DISTRO} -vvvv --config .travis.yml --build scripts
-
-env:
-  - DISTRO="archlinux:latest"
-  - DISTRO="debian:sid"
-  - DISTRO="fedora:latest"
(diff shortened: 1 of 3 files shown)


greg-hellings:
@alerque

greg-hellings:
I'm not sure if we use branches in the repository other than "master". I haven't noticed us do much. If so, we can add additional guard conditions to the deploy so we don't have accidental deployments from branches that shouldn't be released, but for now it seems sufficient to guard on only tags triggering a release.

.ci-build.yml
```diff
@@ -212,14 +181,20 @@ before_scripts:
...
-    cd build
+    set -ex
     export PKG_CONFIG_PATH=/usr/lib/pkgconfig
-    cmake -DCMAKE_INSTALL_PREFIX=/usr -DGTKHTML=ON -DCMAKE_MAKE_PROGRAM=/usr/bin/make ..
-    make -j2
+    cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DGTKHTML=ON -DCMAKE_MAKE_PROGRAM=/usr/bin/make .
+    make -C build -j$(nproc)
+    make -C build mhelp package_source
+    echo "::set-output name=artifact::$(ls build/*.tar.gz)"
+    #if [ "${DISTRO_NAME}" == "fedora" ]; then
```
alerque, line 196: If these cases got taken care of another way, maybe drop the commented code before merger?

.ci-build.yml
```diff
@@ -158,14 +124,17 @@ requires:
...
     - xzip  # Needed to build gtkhtml from source
+    - zip
 
 variables:
   - CFLAGS="-Wall -Werror=format-security"
 
 before_scripts:
   - |
     START_DIR="$(pwd)"
+    WORK_DIR="/tmp"
```
alerque, line 135: I think Actions has an environment variable for where it would prefer you work for scratch stuff. It might be worth reading that in here, which would also be a step towards being able to urn this same build setup locally without hard coding a path that may not be available in all build systems.

.ci-build.yml
```diff
@@ -77,6 +40,7 @@ requires:
     - sudo
     - texinfo
     - which
+    - zip
```
alerque, line 43: This `zip` entry should be above the `# base-devel` comment above as it is a special dependency, not part of the *base-devel* group (which uses `bsdtar` for zip support).

I'm adding this to the AUR package for *xiphos(-git)* as well.

.gitignore
```diff
@@ -9,6 +9,9 @@ src/examples/ipc-interface.xml
 win*/binaries/
 win32/nsis/xiphos*.exe
 win32/po/messages.mo
+cmake/source_version.txt
+xiphos*.exe
+xiphos*.tar*
```
alerque, line 14:
```suggestion
xiphos-*.exe
xiphos-*.tar*
```

A minor thing, but is more specific to the release artifacts (which we expect) and if we ever ended up with a stray *xiphos.exe* in the repo root (which we do not expect) it would show up as not belonging.

.ci-build.yml
alerque: Aaa, from one of your list emails...

> All that remains is adding the Linux artifact stashing and releasing on tag events. That's just a matter of about four more steps in Github [...]

I'm guessing that might be related. Carry on!

alerque:
> I'm not sure if we use branches in the repository other than "master". I haven't noticed us do much. If so, we can add additional guard conditions to the deploy so we don't have accidental deployments from branches that shouldn't be released, but for now it seems sufficient to guard on only tags triggering a release.

I would guard to *master* anyway. When somebody does come along and branch something they will have forgotten (or never understood) this mechanism and it will be a fiasco.

Also looking at the current tag list, I would be inclined to match on tags that are feasible releases as understood by Cmake's numbering system, in other words guard releases to tags matching `^[0-9]+(\.[0-9]+){2,3}$`. Nothing else is going to be properly understood by the release system anyway (and I'm not ever sure about the 4 number variant, since we are setting the .TWEAK version in devel builds based on a commit counter, it might be best to force just a three unit MAJOR.MINOR.PATCH format for releases, so `^[0-9]+(\.[0-9]+){2}$`).

karlkleinpaste:
@greg-hellings do you want to make any further changes as re: above, or should i just merge it?

greg-hellings:
I'm trying out a few of them, so expect this PR to bounce a few more times as I commit and it runs those tests. I'll ping on this thread, again, when I'm done with them.

.ci-build.yml
greg-hellings: I skipped generating these binary artifacts because I couldn't guard them. But I should be able to generate directly with calls to "cpack -G" so I'm testing that now.

.ci-build.yml
greg-hellings: Moved.

.gitignore
greg-hellings: I would disagree on this point. We don't expect it to be there, but "git st" shouldn't be your mechanism for finding things that don't belong. `.gitignore` is just to be sure that it doesn't land in the git tree by accident. I prefer broad git ignore files, because there have been many times I've found stray generated files making their way into the repo that shouldn't have.

.ci-build.yml
greg-hellings: OK, this is now working as expected.

.ci-build.yml
greg-hellings: It does, in fact. ${{ runner.temp }}. Updating the build thusly.

.gitignore
alerque: Sure.

.ci-build.yml
alerque:
```suggestion
    WORK_DIR="${{ runner.temp }}"
```

greg-hellings:
> Not sure if the temp dir thing got missed or you tried and then reverted, not a big deal either way.

I definitely committed only half of the change. Now it's up there.

greg-hellings:
I don't plan to update the .gitignore file here. I prefer it the way that this change has it. And as for guarding the tags, I think that's just fine the way it is. If we guard to just the master branch, then we need to make edits to feature branches in the future. Also, past tags that don't fit semver are from the dark days of SVN and a different world altogether.

alerque:
> I don't plan to update the .gitignore file here.

No problem.

> And as for guarding the tags, I think that's just fine the way it is. If we guard to just the master branch, then we need to make edits to feature branches in the future

You actually **did** guard the release job to tags only, just not where I was expecting it. I usually do it on the `on:` configuration for the workflow, you did it in the job configuration. That's fine. And of course the builds themselves should happen on all commits, branches, and PRs.

> Also, past tags that don't fit semver are from the dark days of SVN and a different world altogether.

August 2017 wasn't dark days of SVN, although it does seem like a different world just since last August.

LGTM.

greg-hellings:
> You actually **did** guard the release job to tags only, just not where I was expecting it. I usually do it on the `on:` configuration for the workflow, you did it in the job configuration. That's fine. And of course the builds themselves should happen on all commits, branches, and PRs.

Proper use of the `if` tag is a newly acquired feature knowledge point for mine. In previous lives I've guarded releases by having separate workflows, entirely, for release. But that meant either skip tests altogether, or duplicate test logic. Neither seemed right. I just stumbled across the `if` conditional recently, and now all my code is getting retrofitted!

alerque:
> Neither seemed right. I just stumbled across the `if` conditional recently, and now all my code is getting retrofitted!

Some of mine will be soon too! A clear advantage for many scenarios, although the way it displays in PRs is a small ding.

PR #1016, opened 2020-05-01
Use pre-built MinGW Biblesync package

greg-hellings:


karlkleinpaste:
@greg-hellings how concerned do i have to be with the unsuccessful checks?

greg-hellings:
@karlkleinpaste None - those are the ones that have been broken in Travis for a while. The fix for those (moving them to Github Actions) is in the other PR.

PR #1020, opened 2020-05-07
Fix up release builds and process:

greg-hellings:
Pull libbiblesync.dll from system installation location
Install biblesync from Fedora package repos
Remove libgsf, which is no longer needed
Allow building the container image to avoid re-downloading dependencies,
over and over again
Don't include new build files in package
Upload both .tar.gz and .zip files
Don't generate RPM or deb files
Simplify logic for source upload, since Github Actions will detect
duplicates, and only the first file will be accepted
v2 of {upload,download}-artifact actions now has greatly streamlined
tools so we don't have to be so uptight about how things are specified.
Move to using this
Update docs

### DIFF
--- a/win32/CMakeLists.txt
+++ b/win32/CMakeLists.txt
@@ -59,9 +59,9 @@ if (WIN32)
       DESTINATION ${CMAKE_INSTALL_BINDIR}
       COMPONENT binaries
       )
-  elseif (EXISTS ${CMAKE_FIND_ROOT_PATH}/lib/libbiblesync.dll)
+  elseif (EXISTS ${CMAKE_FIND_ROOT_PATH}/bin/libbiblesync.dll)
     install(FILES
-      ${SYS_ROOT_LIB}/libbiblesync.dll
+      ${SYS_ROOT_BIN}/libbiblesync.dll
       DESTINATION ${CMAKE_INSTALL_BINDIR}
       COMPONENT binaries
       )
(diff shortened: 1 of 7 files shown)


win32/CMakeLists.txt
```diff
@@ -59,9 +59,9 @@ if (WIN32)
       DESTINATION ${CMAKE_INSTALL_BINDIR}
       COMPONENT binaries
       )
-  elseif (EXISTS ${CMAKE_FIND_ROOT_PATH}/lib/libbiblesync.dll)
+  elseif (EXISTS ${CMAKE_FIND_ROOT_PATH}/bin/libbiblesync.dll)
```
alerque, line 62: This doesn't smell right. A dll in an a bin directory instead of a lib directory?

Then again Windows is a bad smell, so if I'm just out of touch ... carry on.

greg-hellings: All .dll files live in the bin directory for Windows stuff. Windows expects to find the dll files either alongside the file that's importing them or in C:\Windows\system32. There aren't really any other path options. So the mingw framework on Linux tends to do the same. DLL files are placed in ${sys_root}/bin and the .dll.a files (the stubs needed by gcc to link against the DLL) are placed in ${sys_root}/lib instead (there's no way to just point at a DLL and link against it, directly, the way there is for a .so file in Linux). When you distribute to the end user, you package up the stuff in /bin/ and leave the stuff in /lib/ for development installs.

alerque:
Great work @greg-hellings, the artifacts on the Github release look great for the newly minted 4.2.1 tag.

AUR pkgbuild & precompiled Arch packages are up to date using the new sources, and everything seems to have worked like it should.

PR #1060, opened 2020-10-12
Create source_version.txt

dayanandakr:


karlkleinpaste:
pretty sure this is exactly not desirable, as it is dynamically created on purpose, and is mentioned in .gitignore.
@alerque @greg-hellings ?

alerque:
Correct, this should **not*** be included in the project sources. It is generated and cached and put in generated packages, but should not be committed here.

https://github.com/crosswire/xiphos/blob/ece84a44111b790f242ef3aedb520bebd9e977a9/CMakeLists.txt#L60-L62

karlkleinpaste:
that's what i thought. leaving this out.

PR #1061, opened 2020-10-12
Create source_version.txt

dayanandakr:


alerque:
This is a duplicate of #1060 that we just closed and explained why this should not be part of the repository. It does not belong in Git. Perhaps you can explain your usage and why you thought this was necessary, I think you're probably using the packaging and tooling wrong and we could point you in the right direction.

PR #1103, opened 2022-01-15
Add dbus-glib dependency to main

jtojnar:
It is required through the ipc header and the build will fail without it on Nix:

	In file included from /build/source/src/main/search_sidebar.cc:48:
	/build/source/src/gui/ipc.h:26:10: fatal error: dbus/dbus-glib.h: No such file or directory
	   26 | #include <dbus/dbus-glib.h>
	      |          ^~~~~~~~~~~~~~~~~~
	compilation terminated.


karlkleinpaste:
ancient and not needed given current build habits.

PPN-SD:
This patch is still required here (gentoo) for 4.3.2.
Why do you mean by 'not needed given current build habits.' ?

karlkleinpaste:
Pick any recent commit [e.g.](https://github.com/crosswire/xiphos/commit/a61cd4278eaae1828f909a3882cc3f0975fff862), follow its build logs indicated by the green checkmark near top left that says "8/9", which indicates that all 8 test builds ran  successfully, and the 9th stage was not run because it wasn't a release deployment. Click on that indicator and you'll see the 8 ways we build on every commit: win 32 & 64, 2 arch, 2 fedora, 1 debian, 1 ubuntu.

If gentoo still has a problem with it, I have to wonder why -- it's the odd man out. We haven't seen a dbus build failure in eons.

PPN-SD:
But doesn’t it make sense that include_target only applies to the designated target gtk?

PR #1108, opened 2022-03-08
Modifiyng the phrase "Insertar emoticono"

MathewFreddickRiddle:
Because emojis and emoticons are not the same.
An emoticon is formed by text characters, while an emoji is an image.


LAfricain:
@Karl, Maybe we should also change the source text ?

Issue #1083, opened 2021-04-06
Unable to build, "error: template with C linkage"


alerque:
Arch Linux recently updated the SWORD library in the official repos. This makes the unofficial SVN one I've been hosting to provide Xiphos builds obsolete and [the current builds are dead](https://github.com/alerque/aur/issues/15).

In trying to update and for some reason I can't rebuild Xiphos at all. I've tried with both the official SWORD 1.9.0 package and and updated SVN release and got nowhere.

The errors start here and go on *ad nauseam*:

```
[  6%] Building CXX object src/main/CMakeFiles/main.dir/configs.cc.o                                                                                                                                             
In file included from /usr/include/glib-2.0/glib/gatomic.h:31,                                                                                                                                                   
                 from /usr/include/glib-2.0/glib/gthread.h:32,                                                                                                                                                   
                 from /usr/include/glib-2.0/glib/gasyncqueue.h:32,                                                                                                                                               
                 from /usr/include/glib-2.0/glib.h:32,                                                                                                                                                           
                 from /usr/include/gtk-3.0/gdk/gdkconfig.h:8,                                                                                                                                                    
                 from /usr/include/gtk-3.0/gdk/gdk.h:30,                                                                                                                                                         
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,                                                                                                                                                         
                 from /home/caleb/projects/aur/xiphos-git/src/xiphos/src/main/sword.h:28,                                                                                                                        
                 from /home/caleb/projects/aur/xiphos-git/src/xiphos/src/main/configs.cc:28:                                                                                                                     
/usr/include/c++/10.2.0/type_traits:56:3: error: template with C linkage                                                                                                                                         
   56 |   template<typename _Tp, _Tp __v>                                                                                                                                                                        
      |   ^~~~~~~~                                                                                                                                                                                               
In file included from /home/caleb/projects/aur/xiphos-git/src/xiphos/src/main/configs.cc:28:                                                                                                                     
/home/caleb/projects/aur/xiphos-git/src/xiphos/src/main/sword.h:26:1: note: ‘extern "C"’ linkage started here                                                                                                    
   26 | extern "C" {                                                                                                                                                                                             
      | ^~~~~~~~~~                                                                                                                                                                                               
In file included from /usr/include/glib-2.0/glib/gatomic.h:31,                                                                                                                                                   
                 from /usr/include/glib-2.0/glib/gthread.h:32,                                                                                                                                                   
[...]

PR #1134, opened 2023-10-08
Rescope extern C declaration to avoid compile errors

### LINKED ISSUE #1133: Cannot build git master of Xiphos on Fedora 38 - "error: template with C linkage" caused by src/main/sword.h
Attempting to build Xiphos from git master results in a massive number of errors like this:

```
/usr/include/c++/13/type_traits:652:3: error: template with C linkage
  652 |   template<typename _Tp>
      |   ^~~~~~~~
/home/user/proj/xiphos/xiphos/src/main/sword.h:26:1: note: 'extern "C"' linkage started here
   26 | extern "C" {
      | ^~~~~~~~~~
```

The issue can be solved by patching src/main/sword.h with the attached patch. The start of the `extern "C"` block is moved down a bit so that the GTK and GLib includes are no longer inside the `extern "C"` area. This allows Xiphos to build and run successfully.

[0001-Rescope-extern-C-declaration-to-avoid-compile-errors.txt](https://github.com/crosswire/xiphos/files/12778186/0001-Rescope-extern-C-declaration-to-avoid-compile-errors.txt)

ArrayBolt3:
Closes https://github.com/crosswire/xiphos/issues/1133 and https://github.com/crosswire/xiphos/issues/1083.

### DIFF
--- a/src/main/sword.h
+++ b/src/main/sword.h
@@ -22,12 +22,12 @@
 #ifndef _SWORD_H__
 #define _SWORD_H__
 
-#ifdef __cplusplus
-extern "C" {
-#endif
 #include <gtk/gtk.h>
 #include <glib/gi18n.h>
 
+#ifdef __cplusplus
+extern "C" {
+#endif
 #include "main/configs.h"
 
 /*** these defs need to be seen by gui and backend ***/


ArrayBolt3:
The test build failures are unrelated to the code change. There's some version detection logic in the CMake code that fails quite badly if the build is done outside of a Git repo with the proper tag(s). I'll file a bug report about that soon-ish.

Phyrik:
@ArrayBolt3 I can't seem to find any issue regarding the test build failures, was the bug report ever filed?

PR #1139, opened 2024-01-23
Build with webkit2gtk 4.1 and libsoup3

bgermann:
Forwarded from [Debian](https://salsa.debian.org/pkg-crosswire-team/xiphos/-/merge_requests/6):

Debian's webkit2gtk maintainers intend to stop building the 4.0 API soon. The 4.1 API is the same as the 4.0 API except that it uses libsoup3 instead of libsoup2.4.

Fedora has already [stopped](https://fedoraproject.org/wiki/Changes/Remove_webkit2gtk-4.0_API_Version) building the 4.0 API in preparation for the Fedora 40 release in a few months.

libsoup2 -> libsoup3 porting is sometimes [complex](https://gitlab.gnome.org/GNOME/libsoup/-/issues/218). But in this case, I only made a minimal change and I was able to download a Bible translation with Xiphos and things appeared to work ok.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -285,7 +285,7 @@ Create a build directory as a sibling of the xiphos directory:
 
 ## 5. Install the required dependencies:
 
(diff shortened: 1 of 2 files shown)


bgermann:
As far as I can see this stanza will also need to update. I guess the patch submitter just tested with the Debian configuration and therefore only changed the relevant path for that.

karlkleinpaste:
thx. i'll take care of it.

PR #1150, opened 2024-06-02
Fix CI by updating packages

Phyrik:
Current changes only partially fix Debian by resolving missing python-dev package.

Still need to resolve libenchant-dev, candidate to replace it is libenchant-2-dev but will have to change code to work with new version.


karlkleinpaste:
be aware that some changes must be #ifdef'd to preserve windows build using older toolkits that we apparently can't upgrade. i don't remember if libenchant is one of those.

Issue #794, opened 2017-04-24
Update WebKit/WebKit Editor detection


greg-hellings:
Fedora Rawhide (will be Fedora 27) does not have support for WebKitGTK, and requires WebKit2 APIs. Likewise, both Ubuntu has not had gtkhtml since at least 16.04. Both our usage and detection of the webkit2 package needs to be update to look for the new version (webkit2gtk-4.0.pc instead of webkitgtk-3.0) and the WebKit Editor needs to be finished.

karlkleinpaste:
i could use someone willing to wade into what's wrong with the webkit editor for webkit2. i honestly don't understand why the webkit1 API couldn't have been provided exactly by webkit2.

TingPing:
@karlkleinpaste To be frank you are 7 years late to have that discussion.

karlkleinpaste:
@yetist, i see from insights -> network that you have a webkit2-editor branch. have you done work on this? is there hope you will have something we could merge? you've done such useful work in other gtk matters, i'm hoping you've gotten a handle on this as well.

yetist:
Yes, I created the webkit2-editor branch to support webkit2gtk as the html editor, but the API change of webkit2gtk-4.0 is relatively large,  this work is not very easy, it has not been completed yet. If it is done, I will create PR.

jbicha:
I expect that xiphos will be removed from Debian (unstable) and Ubuntu 19.04 later this year. It has already been removed from Debian Buster ("Testing").

It is impossible to build xiphos on Debian Testing because gtkhtml was removed 3 years ago and the old webkitgtk has finally been removed from Testing too.

teusbenschop:
On Debian (unstable) and therefore on upcoming Ubuntu releases, Xiphos was patched in such a way that it now builds without depending on libwebkitgtk. 

That will keep Xiphos in Debian and Ubuntu.

The patches remove the functionality that libwebkitgtk provided. It's about editing the personal commentary.

The patches can be viewed in the source code repository for the Debian packaging:

https://salsa.debian.org/pkg-crosswire-team/xiphos

bgermann:
I propose making the editor component optional, which would also help with #963. If you agree, I am going to work on that myself.

LAfricain:
It's a good idea. But I'm very interresting about it. And some friends are too! The best way would be to change the editor. However in the meantime it would make compiling and creating the deb package under debian much easier if the editor is optional.

greg-hellings:
[...]

PR #1151, opened 2024-06-03
CI fix broken dependencies

Phyrik:
Further commits for the same topic as #1150 

@karlkleinpaste I don't know if you are comfortable with many pull requests for each change I make for each fix, or if you want to hold off on merging until I've got everything sorted, or if this should remain a draft pull request for the time being. Completely up to you.


karlkleinpaste:
since it looks like you'll be changing things for a bit, i'll hold off until you tell me you're at a good endpoint or intermediate point. no sense overdoing it.

Phyrik:
Sounds good, I'll mark this as a draft until I reckon you should merge it

Phyrik:
The problem with Debian and Ubuntu (and likely the other distros, although they have other outstanding issues) is the deprecation of webkitgtk-3.0 (see #794) which is also preventing me from building on my Debian system.

The only reason the package in in the Debian repos is a patch from the maintainer removing the webkit editor (see [here](https://sources.debian.org/src/xiphos/4.2.1+dfsg1-8/debian/patches/0002-Disable-webkit-editor.patch/)).

This seems to be a long standing issue with a lot of work needed to fix but at least it explains the failure of the CI builds.

karlkleinpaste:
just curious, can you tell me where you are with this effort?
i'm trying to get a few things accomplished (notably several PRs completed) and this remains in draft state.
if it needs to wait, that's fine, i just wonder where you are, how far you've gotten.

Phyrik:
Unfortunately I'm unable to make progress until we decide on a route to solve the webkitgtk debacle, since the reason the CI builds are failing is due to missing dependencies on said webkitgtk libraries.

I mentioned this at the start of the email I sent on the xiphos-devel mailing list, so you can refer to it for a more in depth explanation.

If you're happy with taking the route that Debian has (i.e. removing the components, notably the studypad, that require the webkitgtk dependencies) I'd be happy to do so, but it would be a large set of features to remove.

karlkleinpaste:
i did respond to your email to xiphos-devel, and offered some explanation for what i thought i knew. perhaps i was off, but my belief was that webkitgtk was simply moving forward to a most-recent 4.1 version, hence recent commits to xiphos/cmake/XiphosDependencies.cmake, still distinguishable from the old versions that the windows build still needs.

looking at the latest build logs, it seems the fedora and arch failures are strictly on the version stamp:
```
CMake Error at CMakeLists.txt:73 (project):
  VERSION ".128" format invalid.
```

looking at the debian failure, it's unhappy with libenchant-dev; is this now missing, or renamed, or...? ubuntu is complaining about python-dev -- is this needing a version spec, such as perhaps python3-dev (i'm only a very casual ubuntu user)?

sorry if i'm being dense, but i don't see precisely what's gone wrong in this instant, specifically how it relates just to webkitgtk.

if the editor is the real base of the troubles, then #ifdef'ing it out (viz studypad and other editables) as a build-time choice would be ok, and let those features continue where the platform allows (fedora, at least).

Phyrik:
Sorry for the late reply, was away this week just past.

Regarding the Arch and Fedora builds you are completely right that the version number is breaking them. I hadn't focused on them too much, apologies. Unfortunately I have no clue how to solve that, I'm not that familiar with CMake so if anyone has expertise there help would be greatly appreciated.

Debian and Ubuntu are both missing libenchant-dev which is required to do the manual build and install of gtkhtml. So we need some replacement. I tried upgrading it to libenchant2 but that requires changes to the underlying code of the gtkhtml source that is grabbed from the archives (which is a massive hassle and leaves a bad taste in the mouth for any distro packagers).

As for Ubuntu specifically I'll change the CI to use that updated webkit2gtk 4.1 version. The python-dev issue I thought I had fixed, that's a really simple change just need to rename to python-dev-is-python3, will check if there's anywhere I forgot to rename the package.

As for the #ifdefs I'll go ahead with that. You say Fedora wouldn't need it, does Fedora have the old gtkhtml packages in its repos still?

So as for the relation to webkitgtk, it seems to be primarily gtkhtml instead. In all honesty I am still unsure of what the difference in uses is for these two libraries, do they not serve the same purpose? If they do, I would propose removing gtkhtml entirely. It also seems those earlier commits you were referring to didn't change all the references to the new webkit2gtk version, so I'll go about changing the rest of them.

Thanks for your help and support, feel free to correct me if there's anything further you see needs altered.

karlkleinpaste:
fedora continues to have gtkhtml. xiphos may be the last project using it. i think arch also still has it.

we don't use a lot of gtkhtml; it's used only for the editor (libgtkhtml-editor.so), and has for many years been a serious pain in our collective backside. it is desperately in need of being re-written against some other toolkit besides gtkhtml, but no one, including me, has been willing to wade into the swamp. :-/

the version number confusion that causes fedora/arch build failuire has me stymied. i asked several folks privately a couple months ago about it, and sadly no one seems to remember how the overall mechanism works. i didn't implement it; it was part of the cmake conversion when we left `waf` behind, which was handled by others.

your contributions and assistance are hugely appreciated. sorry that interaction here has been slow on both our parts.

domcorbex:
A few notes about the CMake snippet (`CMakelists.txt: lines 63-67`)

The code first checks if a `.git` directory exists in the project source directory:

`if(EXISTS "${PROJECT_SOURCE_DIR}/.git")`

If the `.git` directory exists, it includes a CMake module called `GetGitRevisionDescription`:

`include(GetGitRevisionDescription)`

It then uses the `git_describe function` (provided by the included module) to get the Git version information:

`git_describe(GIT_VERSION --tags)`

The code applies a regular expression to modify the version string:

`string(REGEX REPLACE "-([0-9]*).*" ".\\1" VERSION "${GIT_VERSION}")`

This regex replaces "-<number><anything>" with ".<number>". For example, "v4.2-3-g123abc" would become "v4.2.3".
Finally, it writes the processed version string to the file `cmake/source_version.txt`:

`file(WRITE ${SOURCE_VERSION_CACHE} "${VERSION}")`

$(VERSION) format:
`<major>.<minor>.<patch>[-rc<n>]`
where the <patch> component is less than 20000000. 

To sum up, this CMake code automatically generates a version information based on Git tags. It allows the build system to incorporate the current Git version into the project's version number, which can be useful for tracking builds. 
The release tarball doesn't include a `.git` directory, the version information is then based on the content of `/cmake/source_version.txt`.

karlkleinpaste:
@domcorbex
where/how does the fedora failure arise?
```
CMake Error at CMakeLists.txt:73 (project):
  VERSION ".128" format invalid.
```
there should be no .git directory in (what i presume must be) the generated tarball from which to work. how does it find ".128" -- from where does this come? (cmake/source_version.txt is not in the tree, and is named in .gitignore.)

domcorbex:
IMHO, the script:

1. Checks if the project contains a Git repository.
2. If it is, it reads the version from `git tags`, processes it, and saves it in `SOURCE_VERSION_CACHE`.
3. If it's not, it reads the version from `SOURCE_VERSION_CACHE`

In the case of a tarball, the `.git` folder doesn't exist, CMake reads the version from `SOURCE_VERSION_CACHE` (cmake/source_version.txt). 

If I'm not mistaken, the problem is that `SOURCE_VERSION_CACHE` is only created in the tarball by `cmake package_source`. In the CI, I guess the repo is exported without the `.git` directory, so both the `.git` directory and `SOURCE_VERSION_CACHE` (cmake/source_version.txt) are missing, as a result `VERSION` is undetermined. 

A quick fix would be to create `cmake/source_version.txt` and push the next version string in it.

domcorbex:
I don't know what's going on with CI, the full repo seems to be fetched, with tags and so on.. (.github/workflows/builds.yml)

The latest git cloned on Fedora 40 builds successfully, VERSION=4.2.1.27

Phyrik:
Thanks very much for the guidance @domcorbex ! That's the CMake version stuff sorted out temporarily (currently just sets the version to 0.0.1). We've got some new errors from Arch and Fedora now though.

Arch can't find glib-genmarshal. No clue what this is or what might be a replacement package.

Fedora can't find libsoup-3.0, which seems to be called libsoup3 on Fedora's repos, but I don't know how to tell CMake to differentiate between distros. I'll look into this one.

Phyrik:
I have no clue how the Windows CI builds work, it looks like @greg-hellings created it, so if you are reading this Greg any help would be appreciated!

All the Linux builds are working. If I remove the studypad button from the top bar menu, @karlkleinpaste , would you be happy to merge?

karlkleinpaste:
just to clarify, for those platforms which still provide gtkhtml4-editor, is the editor itself still enabled by this? i was looking through the commits, and the changes in webkit-editor.c give me pause, though i haven't really looked closely at the details yet.

in general, i'm totally ok with merging soon; if you have something remaining to be done, well, just let us know.

@greg-hellings could we convince you to invest what i hope would be no more than an hour or two to re-solidify the windows build mechanisms, including (i suppose) a needed update to the sword patch?

Phyrik:
Ah my bad, I had removed the editor from them all to keep it consistent, I'll re-enable it for those distros that do provide the necessary packages

win32/Dockerfile
```diff
@@ -4,7 +4,7 @@ VOLUME /source
 
 # Installs dependencies once and for all
 RUN dnf -y update --refresh && \
-    dnf -y install cmake git fpc gettext glib2-devel itstool libxslt make yelp-tools zip && \
+    dnf -y install cmake git fpc gettext glib2-devel itstool libxslt libsoup3-devel make webkit2gtk4.1-devel yelp-tools zip && \
```
greg-hellings, line 7: Please note that this is running on Fedora 30. You aren't going to get the latest version of packages on it. For example, I can guarantee you that webkit2gtk4 is not going to be in there, as Fedora 30 reached end of life in 2020. If Xiphos' Windows build cannot be built against the antique versions of the packages found in Fedora 30, then we will need to find another way to do this. The Windows build has been stuck in limbo on very old versions of libraries for a long time because of these limitations. Specifically, though, you will need to install the mingw32/64 versions of these two libraries in order to build.

karlkleinpaste: because there have never been (successful) GTK3 or later packages for Windows, we remain stuck using GTK2. the code has been significantly #ifdef'd for this purpose for many years. (the original Windows port was in late 2009, when GTK2 was in its heyday. we have never moved off of it.)

yes, we know it's ancient and crufty and in some odd sense a security problem, but the use case of xiphos doesn't lend itself to any serious security abuse. it's what we've got to work with, so we build with it.

karlkleinpaste: ...so that means, for Windows purposes, we have to return package qualifiers to the older versions. yes, webkit2gtk4.1 isn't available, and webkitgtk (original for GTK2) requires libsoup2.

greg-hellings: Correct. I was saying this for @Phyrik's edification.

There isn't a libsoup3 build (in Fedora) for MinGW. And the latest webkitgtk seems to be mingw32-webkitgtk3, which is only in very old releases of Fedora hence why we pinned this effort to Fedora 30. No webkit2 at all in the MinGW builds as that was never supported upstream. So the dependencies for Windows will need to specifically work with a version of libsoup < 3 and with the older version of webkit.

greg-hellings: I have pushed up a corrected version of the Dockerfile that does the Windows builds. You can find it on the tip of my branch with the same name as the PR branch here. It doesn't fix the actual build issues, as that is dependent on changes to the C code itself. However, if you follow the two steps in the CI script, you can do the development locally to get it into a good state with those deps.

`docker build -t xiphos win32` from the root of the Xiphos repository will create a local cached version of the build with all the dependencies intsalled. I use `podman` instead of `docker`, and there is no reason you can't, as well, if you prefer. It is the generally better alternative, but is a drop in replacement for you.

After that `docker run -i --rm -v ".:/source" xiphos -win32` and `docker run -i --rm -v ".:/source" xiphos -win64` will test building your local copy against Win32 and Win64 targets (we might want to drop the 32 bit target, eventually, because literally no one has 32 bit architecture these days on Windows). You can use these two commands to iterate on the changes needed to build with libsoup2 and webkitgtk2.

greg-hellings: I'm also not entirely sure you're correct about the GTK still being stuck on 2. I'm pretty sure we have been using GTK3 for the Windows build for quite some time. Additionally [GTK 4 supports Windows](https://docs.gtk.org/gtk4/windows.html). The only thing that prevents Xiphos from using a unified GTK 4 code base across both environments is use of WebKitGTK which dropped all Windows support some time ago and will likely never bring it back without a paying customer to support the effort.

Moving off of WebKitGTK would be a monumental forklift for the project. But, moving forward, it might become necessary. To date, the only really viable options I've seen are the [webview](https://github.com/webview/webview) library and (chromiumembeddedframework)[https://github.com/chromiumembedded/cef]. The former is more lightweight (wrapping existing system libraries on the host) while the latter is more heavyweight (embedded Chromium's rendering engine into your application). Both are well established, integrate with GTK, are widely used, and are intentionally cross platform. WebKitGTK is intentionally not.

I know it is a long-standing question. But the further in time we go, the more it might become important. GTK3 will eventually fall out of support and will stop supporting newer versions of Windows. If we're actually using GTK2 on Windows that will happen even sooner. It has continued to work for Xiphos for a while. But the end of the line is coming for using antiquated toolkits, eventually. And WebKitGTK is our only boat anchor tying us to them.

karlkleinpaste: > And WebKitGTK is our only boat anchor

call me cynical if you must, but when WebKitGTK came along, it was going to the permanent savior of all things HTML-rendered (because of course we _had_ to get rid of xulrunner) and would never be supplanted. so it became the 5th rendering engine known to gnomesword/xiphos.

now we have to look forward to a 6th.

_o frabjious joy._

karlkleinpaste: while i'm at it... in my current windows xiphos 4.2.1.4, installed under wine:
```
$ find -name *webkit*
./bin/libwebkitgtk-1.0-0.dll
./share/webkitgtk-1.0
```
is "1.0" actually gtk3?

karlkleinpaste: it says "WebKitGTK-2.0" right there on the package.
```
$ strings bin/libwebkitgtk-1.0-0.dll | c++filt | grep -i gtk.[123] | egrep -v _imp_gtk_ | sort | uniq
_head_libjavascriptcoregtk_1_0_0_dll
libjavascriptcoregtk-1.0-0.dll
libjavascriptcoregtk_1_0_0_dll_iname
libwebkitgtk-1.0-0.dll
libwebkitgtk-1.0-0.dll.debug
webkitgtk-1.0
WebKitGTK-2.0
```

greg-hellings: That 1.0 refers to the WebKit API that's used, not the GTK version. The current release is 6.0. The naming scheme jumped around a bit at first while they realized they needed to indicate three different versions: the WebKitGTK version, the WebKit API version, and the GTK version it's bound to. 

And I share your frustration with the moving target of engines. I think it was silly for WebKitGTK to drop support and I think it's a shame that there is no way to use Gecko outside of Firefox itself. But I lack the skills and time to maintain a rendering engine, so I'm left to the whims of those who do, sadly.

karlkleinpaste:
on the matter of the version stamp causing build failure...

i have a xiphos tree that was built yesterday sometime. i realized i hadn't installed on the other near-identical machines nearby. so i went into the 1st of these, with this existing, successful build in place, and did `make install`. it failed:
```
# make install
CMake Error at CMakeLists.txt:73 (project):
  VERSION ".128" format invalid.
-- Configuring incomplete, errors occurred!
make: *** [Makefile:1133: cmake_check_build_system] Error 1
```
@greg-hellings can you offer insight into how this occurs, what it means? we see this routinely, every time a commit is made and auto-build is attempted, and workarounds so far here on @Phyrik 's branch have been simply to obliterate the logic entirely for the time being. the error occurs on master branch.

from where does ".128" come? where is it found? this simple pipeline finds nothing:
`grep -r .128 2>&1 | fgrep -v .po | grep -v 'binary file'`

if i re-run the entire build from scratch, then this tree installs on all systems -- because it's then "fresh"?

karlkleinpaste:
Part update report and part status update request...

As you've probably seen, I just made a smallish code change to handle differences required by RHEL9's gcc11. They're small, they didn't hurt recent Fedora, and I hope they don't break anything else. Please feel free to examine, apply, and test on whatever other systems you've got.

Beyond that, I'm wondering where this PR's effort is now. I know I've been struggling toward release for far too long, but I'm hoping we're getting close. I think what's needed is to get back to retaining the editor for those platforms still supporting libgtkhtml-editor and resurrecting the Windows side of the build.

I didn't have to do anything for Sword or BibleSync in RHEL9.

The configuration oddities of RHEL9, requiring libsoup2 when the world has otherwise moved forward to libsoup3, and needing webkit2gtk 4.0 instead of 4.1, represent a problem for cmake/XiphosDependencies.cmake. The use of pkg_check_modules needs to become more sensitive, and I'm not sure how to do this. Would it suffice to add new cmake -D options to specify a platform directly (i.e. -DRHEL9) so that this can then be tested individually? Can this be done overall for other platforms as well, notably Windows' similar need for libsoup2 and older webkit2gtk? I'm not fully comfortable with recommended cmake-isms and hope @greg-hellings  in particular can advise.

karlkleinpaste:
@Phyrik @greg-hellings
I can hardly be one to push too hard, but I am wondering if there is any update.

Phyrik:
Sorry for the long hiatus, started university recently so was a bit swamped. Hoping to get stuck back in to this now though.

Regarding this PR in particular my outstanding issue is with the Windows build on Fedora. I have no clue how to find the mingw versions of the relevant packages to fulfill the build's dependencies. Will have another try some point soon but might need your help @greg-hellings if you are willing to offer it. It has been much appreciated thus far.

greg-hellings:
I will say that, as a general rule, it is considered a poor choice to hard-code a specific thing like one distribution. Unless it is an environment where you control the whole build to binary process and release - such as we do with the Windows release. Doing it for a RHEL-derived distribution would be considered an anti-pattern. Better to discover features than distro names and versions. However, if you are comfortable with maintaining those specific settings for each target environment, then the main reason it is an anti-pattern is the potential maintenance that comes with it. Eventually you end up maintaining a different definition file for each distribution, OS, etc. So deciding that option is really a maintenance decision - are you more comfortable with some more complicated but generic detection algorithms or are you more happy with a simpler mechanism that might require more manual updating each time a new distro version comes out with new versions of libraries, library renames, etc?

So, while best practices might suggest away from having something like `-DRHEL9`, it might be worthwhile for Xiphos to break "best practices" in the interest of maintainability.

As for Windows specific build issues, I think the build is pinned to Fedora 30. It's a very old, now, version of Fedora that last had all the dependent libraries we require for Windows because of webkitgtk dropping support back in like 2.24 for Windows. If the versions of the libraries we need are not available on that distribution, we will need to figure out a way to build them for it. I have, since, learned better ways to pin builds, but for now let's focus on those antique versions of libraries available in Fedora 30 before we try to tackle Yet Another Build Mechanism.

I'll look into what the build failures are here today and post back with what I find.

greg-hellings:
OK, so a little bit of digging here:
webkit2gtk4.0 is for GTK3 and libsoup2
webkit2gtk4.1 is for GTK3 and libsoup3

libsoup3 was released in 2021 and simply won't be available for Fedora 30 (our base for Windows builds), which went end of life in 2020. Also, you won't find it supported in the Windows build of WebKitGTK which was last supported in - I believe - 2.24 released in 2019.

If you want to keep support for Windows builds you will either need to back out support for libsoup3 or maintain support for both libsoup2 and libsoup3 in the builds with the Windows version leveraging the libsoup2 and the old webkit1 support for as long as that is the display kit in use.

The easiest path going forward might be to use the official port of WebKit for Windows, which is based off Cairo. There appears to be a pretty solid way to connect GTK and Cairo together, since GTK has used Cairo to render most (since 2.8) or all (since 3.0) off its control widgets. I don't pretend to be an expert in GTK or Cairo, but the web is replete with examples of how to grab a Cairo pointer that can be used to render onto a widget.

But, for the purpose of this PR, you'll need to find a way to leverage the libsoup2/webkit2gtk4.0 API and libraries for the Windows build.

karlkleinpaste:
I just wanted to mention -- a couple days late, sorry -- that I'm not ignoring this, but I'm in south Florida and the last week has been kinda overly exciting around here. I'll respond some tomorrow, probably.

Phyrik:
This PR can probably be closed. @greg-hellings has sorted this all out in #1194, #1197 and #1198.

PR #1198, opened 2025-02-16
Consolidate build logic

greg-hellings:
Build more things directly in the GitHub Actions
Add more target configurations to the matrix

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -5,7 +5,7 @@ on:
   pull_request: {}
 
 jobs:
-  build_windows:
+  windows:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


LAfricain:
@greg-hellings  I've tried the build locally but it asks for a dependency that isn't in the repositories. I'm running Ubuntu 24.04. I checked in ubuntu_gtk3_wk2_libsoup3  and I didn't see any webkitgtk-3.0 package.

```
- Checking for modules 'gtk+-3.0;webkitgtk-3.0'
--   Package 'webkitgtk-3.0', required by 'virtual:world', not found
CMake Error at /usr/share/cmake-3.28/Modules/FindPkgConfig.cmake:619 (message):
  The following required packages were not found:

   - webkitgtk-3.0

Call Stack (most recent call first):
  /usr/share/cmake-3.28/Modules/FindPkgConfig.cmake:841 (_pkg_check_modules_internal)
  cmake/XiphosDependencies.cmake:134 (pkg_check_modules)
  CMakeLists.txt:101 (include)
```

Phyrik:
Looking at the GitHub actions it looks like the current CI build doesn't look for `webkitgtk-3.0` whereas when I build locally with the current master branch (commit  `3865277c`) it does look for it.

Output of GitHub action https://github.com/crosswire/xiphos/actions/runs/13457931670/job/37605960971 in section build xiphos at line 60 is `-- Checking for module 'gtk+-3.0'` whereas I get `-- Checking for modules 'gtk+-3.0;webkitgtk-3.0'` like @LAfricain.

@greg-hellings first of all thank you very much for the work you've done with the CI to get it working! I presume you've changed the dependencies in CMake somewhere for the CI builds but that hasn't been reflected in the local builds somehow? I thought they would use the same CMake files. I've looked through some of your commits but can't see where the change was, would you be able to point me in the right direction?

greg-hellings:
What is the exact CMake line you are both invoking with? The only time it should be searching for the `webkitgtk-3.0` library is if you are building with GTK 3 but asking for the WebKit1 API. We don't test that combo on Linux because none of the major distributions ship WebKit1 anymore. It was long ago abandoned by WebKit (not just the GTK port).

So yes, you wouldn't find the package available in distributions anymore. And the CMake files we use in CI are the same ones we use when you are doing a local build.

If you share the CMake invocation you are using, I can try to help you out.

Phyrik:
Fixed it now. I was using the line specified in INSTALL.md, but if I instead execute under the source code directory (not the `build` sibling directory) with `cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_MAKE_PROGRAM=make .` (as the CI does) it builds without errors. I'll make a PR for changes to the INSTALL.md file since there's a few other things I noticed that could be fixed in it too.

PR #1199, opened 2025-02-17
Small correction in the Help files, and the French translation

LAfricain:
I've made some minor punctuation corrections, mainly in the xiphos help. I think that, for the moment, the French is correct. I'd just like to point out a few lines that are only displayed in English in the advanced search, such as "[New Range]" and "Sample Module List", and "Old Testament", "New Testament".

If you're working on other bugs with new entries to translate, I'll keep an eye out for them. I'll also take a look at the Italian as soon as possible.


karlkleinpaste:
There are conflicts being noted on this PR. Could you take a look at resolving them, and then I'll merge?

Note that I tweaked a couple things in the "shortcut keys" and the "interface" page (re: comment on parallel vs. v11n differences).

PR #1209, opened 2025-02-24
Dialog alert user if studypad is not supported on their platform

phyrik:
As you know I am new to contributing to Xiphos so if there's anything I've done wrong do let me know!

This simply alerts the user with a GUI dialogue box that their version of Xiphos does not support the Studypad.

### DIFF
--- a/src/editor/webkit_editor.c
+++ b/src/editor/webkit_editor.c
@@ -33,6 +33,7 @@
 #include <glib.h>
 #include <glib/gi18n.h>
 #include "editor/webkit_editor.h"
+#include "gui/dialog.h"
 
 gint editor_create_new(const gchar *filename, const gchar *key, gint editor_type)
 {
--- a/src/editor/webkit_editor.c
+++ b/src/editor/webkit_editor.c
@@ -34,7 +34,10 @@
 #include <glib/gi18n.h>
 #include "editor/webkit_editor.h"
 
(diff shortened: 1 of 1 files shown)


greg-hellings:
You've got some build failures on Arch here, but otherwise I like this idea. This is something I'll let Karl merge, though. I try to avoid merging code in the C code base, as I am much less familiar with it than he is.

Phyrik:
Thanks Greg, I'll chat to Karl about those failures.

==============================

@karlkleinpaste here's the important stuff:
```
/__w/xiphos/xiphos/src/editor/webkit_editor.c: In function ‘editor_create_new’:
/__w/xiphos/xiphos/src/editor/webkit_editor.c:39:9: error: implicit declaration of function ‘gui_generic_warning’; did you mean ‘g_assert_warning’? [-Wimplicit-function-declaration]
   39 |         gui_generic_warning(_("The Studypad is disabled on this platform due to missing dependencies. We hope to re-enable it in a future release."));
      |         ^~~~~~~~~~~~~~~~~~~
      |         g_assert_warning
```
It looks like a missing include but I don't know why it would work on every other distro except Arch. Any ideas?

Phyrik:
Whoops - disregard that, that warning is still present on the other CI builds. Still confused as to how it causes a build failure though. I'll fix that now and see if giving the CI another go fixes itself.

PR #1228, opened 2025-04-04
Build with ICU correctly

### LINKED ISSUE #1227: Module options | Transliteration no longer works in xiphos.exe 4.3.1 (gtk2 webkit1)
Module options | **Transliteration** no longer works in **xiphos.exe 4.3.1** (gtk2 webkit1)

Moreover, applying this option to an **Ancient Hebrew** module causes the direction to switch from **RtoL** to **LtoR**.

What went wrong?

Am I the first to test whether this feature still works?

greg-hellings:
Patch configure.ac to use pkg-config for ICU detection Use --with-icusword in configure arguments rather than hard defining it Don't pass weird arguments to --with-icu

Fixes #1227

### DIFF
--- a/win32/build_sword.sh
+++ b/win32/build_sword.sh
@@ -9,6 +9,7 @@ tar xzf sword.tar.gz
 cd sword-1.9.0
 # It confuses gcc when you have this
 sed -i configure.ac -e 's#-no-undefined##'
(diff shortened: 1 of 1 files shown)


greg-hellings:
@karlkleinpaste This should fix the problem with ICU. At least, I hope it does. If we could get you or David to test this, once the builds are complete, that would be fantastic.

karlkleinpaste:
Thanx. Transliteration is back in Windows.

PR #1242, opened 2025-07-02
fix LTO/One Definition Rule violations

PPN-SD:
>xiphos-4.3.2/src/main/parallel_view.cc:86:15:
>error: ‘no_content’ violates the C++ One Definition Rule [-Werror=odr]
>   86 | extern gchar *no_content;
>      |               ^
>xiphos-4.3.2/src/main/display.cc:88:14: note: ‘no_content’ was previously declared here
>   88 | const gchar *no_content =
>      |              ^

### DIFF
--- a/src/main/parallel_view.cc
+++ b/src/main/parallel_view.cc
@@ -83,7 +83,7 @@ SWBuf unknown_parallel = _("Unknown parallel module: ");
  */
 
 BackEnd *backend_p;
-extern gchar *no_content;
+extern const gchar *no_content;
 
 static const gchar *tf2of(int true_false)
 {


PPN-SD:
> adding `extern` to the instance in `display.cc` causes a different problem.
> 
> ```
> /home/karl/src/bible/xiphos/src/main/display.cc:88:21: warning: ‘no_content’ initialized and declared ‘extern’
>    88 | extern const gchar *no_content =
>       |                     ^~~~~~~~~~
> ```
> 
> it's a good catch except that only `parallel_view.cc` needed a fix to match `const`.

Seems not necessary here indeed.

PR #898, opened 2018-05-25
Add OARS and version tags to appdata

nedrichards:
Adds some more appstream metadata.

I've added basic version tags which help software stores indicate to users what they're about to upgrade to. You can choose to enhance them with real release notes like the contents of https://github.com/crosswire/xiphos/releases if you wish.

OARS is a series of rating tags that lets software stores put location appropriate age ratings. Given what Xiphos ships with I've put a 'there's nothing there' policy in at the moment. You may wish to put in a policy that would seek to rate the contents of the bible as if it was a game:

```  
<content_rating type="oars-1.1">
    <content_attribute id="violence-cartoon">none</content_attribute>
    <content_attribute id="violence-fantasy">none</content_attribute>
    <content_attribute id="violence-realistic">none</content_attribute>
    <content_attribute id="violence-bloodshed">none</content_attribute>
    <content_attribute id="violence-sexual">none</content_attribute>
    <content_attribute id="violence-desecration">none</content_attribute>
    <content_attribute id="violence-slavery">none</content_attribute>
    <content_attribute id="violence-worship">mild</content_attribute>
    <content_attribute id="drugs-alcohol">none</content_attribute>
    <content_attribute id="drugs-narcotics">none</content_attribute>
    <content_attribute id="drugs-tobacco">none</content_attribute>
    <content_attribute id="sex-nudity">none</content_attribute>
    <content_attribute id="sex-themes">mild</content_attribute>
    <content_attribute id="sex-homosexuality">mild</content_attribute>
    <content_attribute id="sex-prostitution">moderate</content_attribute>
    <content_attribute id="sex-adultery">moderate</content_attribute>
    <content_attribute id="sex-appearance">none</content_attribute>
    <content_attribute id="language-profanity">none</content_attribute>
    <content_attribute id="language-humor">none</content_attribute>
    <content_attribute id="language-discrimination">none</content_attribute>
    <content_attribute id="social-chat">none</content_attribute>
    <content_attribute id="social-info">none</content_attribute>
    <content_attribute id="social-audio">none</content_attribute>
    <content_attribute id="social-location">none</content_attribute>
    <content_attribute id="social-contacts">none</content_attribute>
    <content_attribute id="money-purchasing">none</content_attribute>
    <content_attribute id="money-gambling">none</content_attribute>
  </content_rating>
```
Was what I came up with. You may disagree or think about it more with this useful generator: https://hughsie.github.io/oars/

PR #1245, opened 2025-07-16
Add content-rating and releases to appdata

PunkPangolin:
I know you didn't want this in #898, but Flathub now requires both a content rating and release information in the app data file.

Since the app itself does not include any content, I did not add content tags. In that regard, I would argue that it is comparable to a web browser or PDF viewer.

### DIFF
--- a/desktop/xiphos.appdata.xml
+++ b/desktop/xiphos.appdata.xml
@@ -61,4 +61,10 @@
  <url type="homepage">https://github.com/crosswire/xiphos/</url>
  <update_contact>xiphos-devel@crosswire.org</update_contact>
  <translation type="gettext">Xiphos</translation>
+ <content_rating type="oars-1.1" />
+ <releases>
+   <release version="4.3.2" date="2025-04-10">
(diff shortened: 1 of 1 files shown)


greg-hellings:
> > Is there any way to make at least the value of the release attribute dynamically inserted by the build process?
> 
> This should be possible (right now the build seems to add translations into the file?), but I don't know how to write CMake builds; I only stumbled upon this when updating the Flatpak and wanted to upstream the fixes.
> 

We would want to tie in with that CMake process, so that we do not need to manually update the version there when a new release is cut.

> > Also, what is the meaning of "oars-1.1"?
> 
> "OARS-1.1" rates the content of the application according to the [Open Age Ratings Service](https://hughsie.github.io/oars/). This includes tags for depictions of violence, for example. The Bible itself contains much of the content that should be tagged according to OARS, but I am unsure whether that applies to the app. In the one-liner in the PR, no content is tagged.

Ah, I see, that's just saying any content tagged would be using this version of the spec. I misinterpreted that as saying the oars-1.1 was an actual rating itself. Got it.

PR #1263, opened 2026-01-24
Fix HTML tag in Dutch translation for personal area

jajpater:
Small change but not patching it makes a fresh build of Xiphos on NixOS crash.


jajpater:
Fixes #1266

Issue #1104, opened 2022-01-21
Add two new entries in the settings to choose different morphology dictionaries


LAfricain:
Currently the only morph dic we can use is Parkard and Robinson in English. I translated this two dic in French, and it would be nice to have this features in the settings, as it is already possible to chose a different strong dic. This option is already in Bibletime.

![settings](https://user-images.githubusercontent.com/19690477/150501212-3e71d4dc-f819-4789-ad3d-68afce2318ff.png)

LAfricain:
@nwalkewicz do you think this issue is in your skill? I don't think it's too much work, but it's a pretty useful option.

nwalkewicz:
I don't know yet (still pretty new to the codebase, C, and GTK), but I'd love to give it a shot! I'll write here again (with a mention) after the weekend when I have some more time to look into it.

nwalkewicz:
@LAfricain Unfortunately I think this one is beyond me, for now at least; I've actually had trouble getting dictionaries to work at all in Xiphos.

LAfricain:
I even wonder if it wouldn't be useful to have an entry for the NT and one for the OT. As, it seems, there are two different dictionaries.
The real problem here is that without this entry it's impossible to choose a dictionary other than English. Which still limits the possibilities. I've produced a French version of Robinson, and for the time being it's not usable under Xiphos, except with some tweak like renaming it in the conf.

karlkleinpaste:
Not ignoring this. Planning to take a look in the next week. But I have travel obligations that will keep me very busy, from end of February through all of March.

Issue #1010, opened 2020-04-19
Add a new tab in the dictionary pane for daily devotional


LAfricain:
I propose to add a new tab in the dictionary pane, like in the commentary pane. In the commentary pane we have two tabs: commentary and book.
If we have the same in the dictionary pane one for dictionary and one for daily devotional it will be easier to change quickly between the dic and the daily devotional. 
For now I need always to go in the list of module...

DavidHaslam:
Seconded!

Issue #732, opened 2014-05-23
Back and forward buttons in the dictionary view


xiphos-bug-inhaler:
Back and forward buttons in the dictionary would be really useful e.g. after clicking on strongs links for related words.

Reported by: hoodbild

Original Ticket: [gnomesword/bugs/195](https://sourceforge.net/p/gnomesword/bugs/195)

AlisterH:
Have you closed this as "won't fix", or what?
Thanks.

karlkleinpaste:
Um... hm, oops, looks like this got inadvertently closed as a side effect of the bug/enhancement inhale from sourceforge.  re-opening for future cogitation.

PR #1273, opened 2026-03-02
Addition of arrows for history in the dic window, tab for the daily devotional and 3 entries for morphological dictionaries. Solves #1010 #1104 #732

LAfricain:
3 entries added to preferences: #1104 

- An entry for Hebrew morphology
- An entry for Old Testament Greek morphology
- An entry for New Testament Greek morphology
- Resolution of the initial problem of persistent choices

Addition of a daily devotional tab in the dictionary window: #1010

- Display of the dd-specific css
- Display of the date according to the region
- Opening of a calendar for the choice of the day when clicking on the date

Addition of an arrow to navigate through the dictionary window history (#732 )

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -26,8 +26,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Xiphos 4.0.7\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-02-24 22:16+0100\n"
-"PO-Revision-Date: 2026-02-24 22:22+0100\n"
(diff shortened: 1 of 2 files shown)


LAfricain:
After investigation, I can confirm that PR #1273 is NOT redundant and has not been previously merged into upstream/master, you can merge it.

The commits it contains are unique and address three distinct issues:
- #1010: new tab for daily devotional with CSS support and calendar display
- #1104: three new preference fields for morphological dictionaries
- #732: addition of history navigation arrows in the dictionary window

Issue #911, opened 2018-08-14
Font preferences dialog has UI issue


alerque:
Under _Preferences_ → _Fonts_ → _Sizes and Fonts_ → _Font preference, per language_, there is a problem with the drop down menu. I have two languages active, so the menu has "--select--" plus entries for my two language (EN + TR). If I change the selection in the menu I get the dialog promoting for a font choice as expected. However if the language I want to change is already the selected entry, there doesn't appear to be any way to get the font prompt again. The only work around is to select a different language (or the "--select--" entry works too) and then change back to the language I wanted to pick for in the first place.

There should be some way to trigger this prompt directly, or better yet the font preferences should be directly shown in the main dialog, one per language or something like that. The current UI is not just clumsy, it actually feels broken because you can't intuitively get back to the value you just set.

Using Xiphos version `4.1.0-38-gf965eaeb`.

karlkleinpaste:
i know exactly what you mean, and i've thought about it before myself.
problem is that changing the pulldown setting is what sends the gtk signal that induces action.
i honestly don't know how to get the signal sent without changing the state of the pulldown.

alerque:
How about changing the interaction so that the pull-down menu just makes the section and not actually trigger the action, and add a button next to it for "change". That would be much clearer usage and alleviate the current catch 22.

Better yet embed the module font selection in the same UI. Having modal dialogs popping up on top of modal dialogs is confusing and cluttered anyway, there is plenty of room in the Preferences UI to just put those fields below the language selector.

karlkleinpaste:
> pull-down menu just makes the section and not actually trigger the action, and add a button next to it for "change".

that would make xiphos' pulldowns work differently than every other pulldown on the planet, i think. shall we also make the BCV pulldowns of the main window use a "do it now" button? users expect changing pulldowns to induce action on the spot.

greg-hellings:
I dunno, I'm familiar with both Apply/Cancel/OK and immediate reaction UXes. Especially on the web, but I'm also familiar with them in desktop apps.

karlkleinpaste:
solved by #1279

PR #1279, opened 2026-03-08
Fix font preferences dialog UI issues (#911)

LAfricain:
## Changes

- Reset language selector to "-- Select --" after each selection,
  allowing the font dialog to be triggered again for the same language
- Localize "Cancel" and "OK" buttons in font dialog
- Localize "default" in display columns combobox
- Localize "language: %s" prefix in font preferences with lowercase and
  added space after colon, as it is the second part of the sentence
  "Change font for / language: %s"
- Localize "none" in font preferences (utilities.c)

## Files changed

- `src/gtk/preferences_dialog.c`
- `src/gtk/font_dialog.c`
- `src/gtk/utilities.c`


LAfricain:
Hi Karl,

>Do I understand that you've made the dictionary pane now a notebook containing dictionary and devotional, much as the upper right pane is commentary and genbook?

Absolutely! And I must admit, it's extremely comfortable. I hope you like it!

>A (minor?) concern is over the "language" changes.

Thank you for pointing this out. You're right! Fortunately, it's not too difficult to change.

We can revert that change and keep the original "Language:%s" format for the key construction in preferences_dialog.c. I'll do it.

 However, I know it's a minor detail, but could we find a way to localize it?
If we want to display a localized label in the font dialog, we should handle that separately at the display level (e.g. in the label_mod widget in font_dialog.c), without touching the key itself.

Sorry for the oversight!

LAfricain:
> One more test: I re-installed the .24 build, saved its prefs.glade, installed the new .28 build, replaced its prefs.glade with the saved .24 prefs.glade, and once again Prefs dialog opens. The problem is definitely an error in the new prefs.glade.

So, are you informing me so I can handle it, or are you taking care of it?
I'm working on parallel view.

PR #1289, opened 2026-03-14
editor: fix BOOK_EDITOR navigation and content loading

LAfricain:
- Fix widget reparenting for hpaned layout (deprecated gtk_widget_reparent)
- Add html_widget alias for treekey-editor.c compatibility
- Clear buffer before loading new content in _load_text_into_buffer


LAfricain:
@karlkleinpaste , I just thought of something— could you clean up the code from the old editors? Also, the install.md file isn't completely up to date.

PR #1292, opened 2026-03-15
display: implement DisplayLevel for genbooks (fix issue #528)

### LINKED ISSUE #528: Xiphos ignores DisplayLevel option in genbooks
Xiphos behaves always like DisplayLevel=1, even tough it is set to something else.

ps. please move to github ;)

Reported by: tuma3

Original Ticket: [gnomesword/bugs/529](https://sourceforge.net/p/gnomesword/bugs/529)

LAfricain:
Xiphos was always behaving as if DisplayLevel=1 for genbooks, regardless
of the value set in the module configuration. This meant that clicking on
a node in the tree would only display the content of that single node,
with no way to aggregate child content into a single scrollable view.

This fix reads the DisplayLevel config entry from the module and, when
it is greater than 1, recursively renders the current node and its
descendants up to the specified depth, concatenating their content with
a horizontal rule separator between each node.

Changes:
- display.cc: add #include <treekeyidx.h>
- display.cc: add static helper _render_display_level() which traverses
  the TreeKeyIdx tree by offset, rendering each node's content into a
  SWBuf, up to max_level depth
- display.cc: in GTKEntryDisp::display(), read DisplayLevel from module
  config and call _render_display_level() when value > 1, falling back
  to the original single-node rendering when DisplayLevel is absent or 1

Module maintainers who want to benefit from this feature need to add
DisplayLevel=N to their module's .conf file, where N reflects the depth
of the tree they want rendered on a single click.

Fixes #528

### DIFF
--- a/src/main/display.cc
+++ b/src/main/display.cc
@@ -24,6 +24,7 @@
 #endif
 
 #include <glib.h>
+#include <treekeyidx.h>
 
 #include <osisxhtml.h>
 #include <thmlxhtml.h>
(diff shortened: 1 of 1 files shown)


src/main/display.cc
```diff
@@ -725,7 +726,7 @@ CacheHeader(ModuleCache::CacheVerse &cVerse,
 	    GLOBAL_OPS *ops, BackEnd *be)
 {
 	int x = 0;
-	gchar heading[8];
```
phyrik: Instead of increasing the size of the buffer here, would it be better to change the uses of sprintf to snprintf (which takes a maximum number of characters to print)?

LAfricain: Like that:
`sprintf(heading, "%d", x);`
by :
`csnprintf(heading, sizeof(heading), "%d", x);`

phyrik: It would be `snprintf(heading, sizeof(heading), "%d", x);` (without the c at the start) but yes that's the right idea. I think it needs done on lines 735 and 750.

phyrik: For some reason I cannot see those warnings on my build. What command produces that output for you?

karlkleinpaste:
FYI I'm avoiding merging this one until this question is resolved.

PR #1305, opened 2026-04-16
Fix inaccurate CMake dependency comments and downgrade FATAL to WARNING

colelanders:
The CMake dependencies in XiphosDependencies.cmake contained a few inaccuracies:

- A few comments indicating WebKit1 vs GtkHTML editor support were swapped
- The `NOT WEBKIT1 AND NOT GTKHTML` configuration used `message(FATAL ...)` which does not actually stop the build, and the comment incorrectly stated this configuration does not build. 

As of testing on Fedora 43, with the flags `-DGTKHTML=OFF` and `-DWEBKIT1=OFF`, the configuration builds successfully but without editor support. Downgraded to message(WARNING ...) and updated the comment to accurately reflect the actual behavior.

### DIFF
--- a/cmake/XiphosDependencies.cmake
+++ b/cmake/XiphosDependencies.cmake
@@ -121,7 +121,7 @@ if (GTK2)
     )
 else (GTK2)
   if (WEBKIT1 AND GTKHTML)
-    # Gtk+-3.0 + Webkit1 + WebKit-editor
+    # Gtk+-3.0 + Webkit1 + GtkHtml-editor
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
Sorry, this one I'm rejecting.
The editor is back in Linux builds since March 14 commit f5cb5723, using -DGTKTVEDITOR=ON, otherwise it's a WK2 build.
If you'd care to adjust the documentation to reflect that, it would be most welcome.

Issue #1300, opened 2026-03-25
Preview Pane visable but seems broken


asciiman0:
Version 4.3.2 on Linux Mint Cinnamon 22.3

The preview pane on left side is no longer loading the daily devotion.
That pane seems generally broken as it does not seem to load anything.
I am unsure if it is a Cinnamon window management issue or a change in Xiphos.

<img width="1101" height="779" alt="Image" src="https://github.com/user-attachments/assets/5b71fd38-df69-487b-83e5-5de31f0a3508" />

All non-canon related modules now appear in the Dictionary/Devotional window.
The Preview pane seems to be rendered useless.

LAfricain:
I can confirm this bug; it's likely due to the addition of the tab... Thanks for reporting it. It looks like you're using the PPA version.

LAfricain:
@Karl, I won't be available until May 1st, and where I am right now, I don't have access to Claude. So I won't be able to debug this.

asciiman0:
> It looks like you're using the PPA version.

I believe this is from the Mint repository.  Mint used to also have the Flatpak version available in their Software Mgr app.  But today only the 'system package' is available, which seems to work better for me anyway.

<img width="800" height="867" alt="Image" src="https://github.com/user-attachments/assets/ad4290a1-65f7-453f-8ba2-68abbfb8823f" />

LAfricain:
Yes you use the ppa. I updated it with many new feature, This bug is probably a collateral damage.

phyrik:
What is the expected behaviour here? What should be appearing in the preview pane? (I have not used the daily devotional feature so forgive me if I am missing something obvious here.)

LAfricain:
No problem! The preview have to show the references of the day. Only this.

Le 29 mars 2026 16:12:52 GMT+01:00, Scott Buchanan ***@***.***> a écrit :

phyrik:
I'm on the master branch and when I click on the reference in the devotional pane it does show in the preview pane, is it supposed to show automatically or is this a bug on some other version of the code?

asciiman0:
In all my previous versions, once I have set it up, the Daily Devotional will load at launch in the preview pane.  Nice little feature!

phyrik:
Did that happen by default or was there a setting you changed to get that behaviour? I can make it open in the devotional pane on launch but not the preview one.

asciiman0:
Under the 'View' menu:  'Preview Daily Devotional.'
[...]

PR #1308, opened 2026-05-03
Fix daily devotional preview, Restore the ability to manually enter a date in the Devotional tab on Windows

LAfricain:
* Restore checkable 'Preview Daily Devotion' menu item; save and restore state across sessions.
  * Restore the ability to manually enter a date in the Devotional tab on Windows without losing the calendar on Linux.
Solves #1300

### DIFF
--- a/src/gtk/dictlex.c
+++ b/src/gtk/dictlex.c
@@ -619,30 +619,38 @@ GtkWidget *gui_create_devotional_pane(void)
 	gtk_widget_show(hbox);
 	gtk_box_pack_start(GTK_BOX(box_devot), hbox, FALSE, FALSE, 0);
 
-	widgets.entry_devotional = gtk_entry_new();
+widgets.entry_devotional = gtk_entry_new();
 	gtk_entry_set_max_length(GTK_ENTRY(widgets.entry_devotional), 5);
-
+#ifdef _WIN32
+	/* Windows: GtkCalendar is broken — show the MM.DD text entry instead */
+	gtk_entry_set_width_chars(GTK_ENTRY(widgets.entry_devotional), 6);
+	gtk_widget_set_tooltip_text(widgets.entry_devotional, _("Date MM.DD"));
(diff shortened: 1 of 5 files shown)


karlkleinpaste:
Did a test build of the patch to see what's up.

The problem I have with this now-checkable `Preview Daily Devotional` is that it was not previously a checkable item in the first place. It had no effect on `Prefs -> General -> Options -> Show Daily Devotional at start`. It was nothing more than an in-this-instant "show today's daily devotional," right now, regardless of whether that option was set in `Prefs`.

Load up a pre-devotion-in-tab version of Xiphos and see what it does.

Or build an older version yourself to see. I suggest `git reset --hard 4a52a7318a` as a fairly arbitrary point in the semi-distant past (relatively speaking) before the devotional-in-tab came to be.

LAfricain:
I'm not sure we're talking about the same thing. I was trying to fix what wasn't working anymore, namely the preview of the daily devotional found in the View menu under “Preview Today's Daily Devotional.” That's all. I had even forgotten that it could also be checked in the Preferences. So this PR just fixes that (and the calendar Windows issue about choosing days). The only difference is that since we now have a dedicated tab, the preview appears in the tab instead of the preview window.
So, I’m just wondering if it makes sense to have two places to make this selection? Maybe you could remove one of them—either from the View menu or from Preferences?

karlkleinpaste:
The patch was in 2 parts. The 1st did the bugfix (very useful), the entirety of the 2nd was to turn the main menu's `Preview Daily Devotional` into a checkable item. So I made a separate change using just the 1st part, and dropped the 2nd.

Whether we should have a `Preview Daily Devotional` in the main menu, as well as the dict/devo tab selection... Well, one could argue either way. For now, I'm leaving the main menu item in place, which now has the same effect it has always had, which is to put the devotional into whatever-space-it-occupies, which is now the devotional tab.

PR #1314, opened 2026-05-19
Fix shell command safety: replace popen() with g_spawn functions

lggcs:
Replace two popen() calls with GLib's g_spawn_async() and g_spawn_sync() to avoid passing arguments through a shell. This prevents filenames or other data containing shell metacharacters from being interpreted as shell commands.

- src/main/url.cc show_separate_image(): use g_spawn_async() with an argument array and G_SPAWN_STDOUT_TO_DEV_NULL | G_SPAWN_STDERR_TO_DEV_NULL flags, matching the original redirect-to-devnull behavior. Report errors via GError.

- src/gtk/menu_popup.c on_rename_perscomm_activate(): use g_spawn_sync() with sed invoked via argument array. Capture stdout and write the result via g_file_set_contents(). Check exit status and GError.

### DIFF
--- a/src/gtk/menu_popup.c
+++ b/src/gtk/menu_popup.c
@@ -1354,7 +1354,6 @@ G_MODULE_EXPORT void on_rename_perscomm_activate(GtkMenuItem *menuitem,
 	const char *conf_old;
 	char *conf_new;
 	char *sworddir, *modsdir;
-	FILE *result;
 
 	// get a new name for the module.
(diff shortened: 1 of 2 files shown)


lggcs:
@karlkleinpaste If you have a private way of contacting you I can share additional information about this fix and impact. Thank you.

karlkleinpaste:
@lggcs my email is my name k...@k..........org, fill in the dots.

PR #1320, opened 2026-06-14
Addition of all missing languages in the module manager

LAfricain:
Added all languages classified as “Unknown” to the module manager list. This includes Crosswire, IBT, StepBible, and ebible.org, among others.
@karlkleinpaste, Could you check if the issue with the Coptic line is still ongoing? I added an entry for cop-sah, but I didn't dare to modify the Coptic line... The line that should be in Coptic characters.

PS: the build for Windows fails because crosswire.org is down.

### DIFF
--- a/ui/languages
+++ b/ui/languages
@@ -2306,6 +2306,7 @@ eja	Ejamat
 eka	Ekajuk
 eke	Ekit
 ekg	Ekari
+ekk Eesti Keel
 eki	Eki
 ekm	Elip
 eko	Koti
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
I don't see the point of all the blank lines. This isn't a file that needs to remain readable, it's literally just a pile of data not normally seen by humans at all. The few comments after the first 500 or so items (which are historic from very early in then-GnomeSword's life) are to leave observations about problematic or conflicting instances.

I'm a bit mystified how this patch can say "no conflicts with base branch" when the "ekk" entry is already in the latest version in the repo. This would re-add it, and again with a SPC, not a TAB. But it's part of this patch, around line 18 and in the 70s. That syntax (ABBREV TAB NAME) must be carefully adhered to.

Coptic is no longer a problem. I don't recall why this was a problem way-back-when; it was probably some font rendering glitch in then-current gtk. I will correct that.
screenshot11

PR #1324, opened 2026-06-16
fix: prevent stack-buffer overflow on malformed URLs in main_url_handler

hyder365:
In `main_url_handler()`, when a URL containing `passagestudy.jsp` or `xiphos.url` lacks a `?` character (i.e., has no query string), `strchr(url, '?')` returns NULL. The subsequent `++place` triggers undefined behavior, and `strncpy` writes with an effectively random length, overflowing the stack buffer `tmpbuf[1023]`.

Added a NULL check for `place` before use. If no `?` is found, the allocated `GString` is freed and the handler returns early.


karlkleinpaste:
Excellent. I very much appreciate this sort of support.

PR #1325, opened 2026-06-16
fix: prevent URL parameter injection in user annotation links

hyder365:
User annotation content (`e->annotation->str`) was being interpolated directly into the `value=` query parameter of the `passagestudy.jsp` URL in both `GTKChapDisp::display()` and `DialogChapDisp::display()` without URL encoding.

While basic HTML characters like `<`, `>`, `"` were escaped, URI-significant characters such as `&`, `=`, `#`, `+`, and `%` passed through unmodified. A crafted annotation like `&module=EvilMod&passage=EvilRef` could inject arbitrary URL parameters.

Applied `g_uri_escape_string()` to the annotation value before embedding it into the href attribute in both locations.

### DIFF
--- a/src/main/display.cc
+++ b/src/main/display.cc
@@ -1610,13 +1610,16 @@ GTKChapDisp::display(SWModule &imodule)
 
 		// insert the userfootnote reference
 		if (e) {
+			gchar *escaped_value = g_uri_escape_string(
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
Thanx much for the fix.

PR #1327, opened 2026-06-18
fix: replace fixed 1023-byte stack buffer with GString in main_url_handler

hyder365:
The URL handler for passagestudy.jsp-style URLs used a fixed 1023-byte
stack buffer (tmpbuf) with strncpy to extract the portion of the URL
before '?'. When the URL prefix before '?' exceeds 1022 characters,
strncpy fills the buffer without null termination, and the subsequent
null-byte write at tmpbuf[place - url] writes past the boundary.

Replace with dynamic GString allocation via g_string_append_len.

Commit adf7ae7f upstream added a null-pointer check for the result of
strchr(url, '?'), which is important but does not address the
fixed-buffer overflow. This fix replaces the stack buffer entirely.

### DIFF
--- a/src/main/url.cc
+++ b/src/main/url.cc
@@ -886,7 +886,6 @@ gint main_url_handler(const gchar *url, gboolean clicked)
 
 		// another minor nightmare: re-encode / and : in hex.
 		gchar *place;
-		gchar tmpbuf[1023];
 		GString *tmpstr = g_string_new(NULL);
 
 		place = (char *)strchr(url, '?'); // url's beginning, as-is.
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
This one has conflicts. Looking at how the code at now stands, with the other 4 PRs complete (did I cause the conflict by completing others out of order?), I'm a bit mystified for what to do with this. If you could take a 2nd look and adjust it, I'll be happy to complete it.

hyder365:
Rebased this branch onto current master and resolved the conflict. The updated diff keeps the missing-`?` guard from #1324 and preserves the showStudypad/showImage validation from #1328; the remaining change only removes the fixed 1023-byte stack buffer and uses `g_string_append_len` for the URL prefix.

PR #1333, opened 2026-06-23
Reject malformed passagestudy URLs

hyder365:
Malformed `passagestudy.jsp` URLs can reach `main_url_handler()` from remote-control paths such as D-Bus. If required query parameters are missing, the dispatcher can pass null values to `strcmp()` or action handlers.

This change validates the action and required parameters before dispatch, returns failure for malformed action URLs, and avoids dereferencing a missing module entry for user-note previews.


karlkleinpaste:
Again, much appreciation for your analysis and fixes.

PR #1335, opened 2026-06-28
popup menu for multi-reference bookmark navigation (#crossref)

LAfricain:
Add a user-controlled option for navigating multi-reference bookmarks:

- When a bookmark key contains multiple references (semicolons, commas, or verse ranges), the user can choose between two behaviours via a new 'Open cross-references in a popup menu' toggle in the bookmark context menu.

- When enabled (checked): a popup menu lists each reference individually; clicking one navigates directly to that verse. Verse ranges (e.g. Mt 15:1-9) are displayed with their full label but navigate to the first verse of the range.

- When disabled (unchecked): the full key is passed to main_url_handler which displays all references in the Verse List as before.

- Single references and studypad bookmarks are unaffected.
- The preference is saved to and restored from the Xiphos config file under misc/crossref_popup (default: 0 = verse list behaviour).

### DIFF
--- a/src/gtk/bookmarks_menu.c
+++ b/src/gtk/bookmarks_menu.c
@@ -286,6 +286,31 @@ G_MODULE_EXPORT void on_allow_reordering_activate(GtkMenuItem *menuitem,
 
 /******************************************************************************
  * Name
+ *   on_crossref_popup_activate
+ *
+ * Synopsis
+ *   #include "gui/bookmarks_menu.h"
+ *
+ *   void on_crossref_popup_activate(GtkMenuItem *menuitem,
+ *                                   gpointer user_data)
+ *
+ * Description
+ *   toggle popup menu for cross-references
+ *
+ * Return value
+ *   void
+ */
+G_MODULE_EXPORT void on_crossref_popup_activate(GtkMenuItem *menuitem,
+                                                gpointer user_data)
+{
+    settings.crossref_popup =
+        gtk_check_menu_item_get_active(GTK_CHECK_MENU_ITEM(menuitem));
+    xml_set_value("Xiphos", "misc", "crossref_popup",
+                  settings.crossref_popup ? "1" : "0");
+}
+
+/******************************************************************************
+ * Name
  *   on_dialog_activate
  *
  * Synopsis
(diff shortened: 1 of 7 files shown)


karlkleinpaste:
This seemed good. I will make one adjustment to provide for initialization in xml.c.

karlkleinpaste:
Oh no. I did this too soon.

Glitch. Big glitch.
This is what a verse dump of mine looks like in the old way, without popup.
Screenshot at 2026-07-04 11-01-51

This is what the same verse dump looks like in the popup.
Screenshot at 2026-07-04 11-03-21

If I click the entry for Acts 8:36, being fully specified, it goes there correctly.
But when I now click the entry for any other partially specified item, it goes ... somewhere utterly random.

The problem is that incomplete specification is handled properly when the verse list is processed by Sword.
That is, this bookmark entry that I handcrafted...
```
<Bookmark modulename="" key="Matt.17.20; 18.10; 23.13; Mark.7.15; 9.43,45; 11.25; 15.27; Luke.17.35; 23.16; John.5.3; Acts.8.36; 15.33; 24.6; 28.28; Romans.16.23" moduledescription=" " description="Verses should not exist"/>
```
...works fine when processed by Sword because it contextually fills in the fact that e.g. "18:10" is part of a Matthew citation. So it fills that in when returning the list of individual verses.

The popup handling does not provide for this. It leaves the partial (but proper and normal) specification as partial, leaving it contextually unanchored when referenced alone.

I don't want to revert this because the concept is fine, especially being user-selectable so the old behavior is default. But it very badly needs a fix to pass the verse list through Sword for contextual resolution.

LAfricain:
Ok this week I'm busy busy!! Next week I will look at this.

karlkleinpaste:
No sweat. The sole problem is that you still need to pass the bookmark's verse list through Sword just as is done when preparing the verse list dump in the default no-popup case.

LAfricain:
In the end, Claude solved it for me in 15 minutes! I hope that's okay, but take a look at the screenshot—it seems to match the bookmarks you gave me.
crossref

karlkleinpaste:
We have a semantic conflict happening here. I didn't want to file a separate issue until you'd seen this.

I have a bookmark (you have it in my stripped bookmarks.xml):
```
<Bookmark modulename="" key="Matt 15:6; Mark 7:9; Matt 15:1-9; Mark 7:1-13" moduledescription=" " description=""/>
```
When selected, this produces (in both popup and old non-popup) the limited set:


The range portions X:M-N have been reduced to just their 1st verse. Prior to this work in bookmarks, that would have expanded to the full set. How/why was this reduced to just the leading verse? I actually found it very useful to get the entire set.

Issue #968, opened 2019-11-01
Add the possibility to create tags in text


LAfricain:
The idea is very simple, it is to do in Xiphos what [Ezra project](https://github.com/tobias-klein/ezra-project) does.
That means, create tags with Ref of the Biblical text.

karlkleinpaste:
i'm not going to start reading from scratch about yet another bible project.
context: what kind of tags, where placed, how referenced, and in what way useful.

LAfricain:
Ok I'll come back with more information.

LAfricain:
Please reopen this issue.

LAfricain:
Thanks, Karl,
Have you had a chance to read my [comments](1296#issuecomment-4365713990) on the PR? I really don't want to pressure you, but I have to admit I'd be happy to move forward on these issues so I can work on something else, because despite everything, it seems to me we're close to finishing this one.

LAfricain:
@karlkleinpaste,
So that we can make progress on this, I suggest we try splitting my changes into several commits (if possible). Let me know which ones you approve, and which ones you’d like to hold off on or reject. I see three different things:
* The bookmark color
* The highlighting of verses
* The cross-reference context menu

I think the first two work correctly, so we could approve them. You don’t agree with the last one—could we discuss it?

PR #1336, opened 2026-06-28
Highlight verses using tag folder color. Solves #968

LAfricain:
## feat: highlight verses using tag folder color

This PR adds visual highlighting of Bible verses based on their bookmark tag color.

### New behavior

When a verse is bookmarked inside a colored tag folder, it is highlighted in the Bible text using that folder's color. The text color is automatically adjusted (white or black) based on the background luminance to ensure readability.

- The current verse green highlight is disabled when a tag color is present, avoiding visual conflicts.
- Verse ranges, comma-separated verses and semicolon-separated references are all supported.

### Opting out

This feature is entirely driven by the tag color. If you do not want a tag folder to highlight its verses in the Bible text, simply remove its color (click **"No color"** when editing the folder). Folders without a color will not highlight any verses.

### Details

- Highlighting is handled in `display.cc` via `get_tag_color_for_versekey()`
- Text color contrast is computed using the ITU-R BT.709 relative luminance formula
- Depends on the bookmark color feature introduced in PR #1317

### DIFF
--- a/src/backend/sword_main.cc
+++ b/src/backend/sword_main.cc
@@ -925,11 +925,31 @@ GList *BackEnd::parse_verse_list(const char *module_name, const char *list, char
 	key->setText(current_key);
 	vs = key->parseVerseList(list, *key, TRUE);
 
-	if (!vs.getCount())
+	int count = vs.getCount();
+	if (!count)
 		return retlist;
-	while (!vs.popError()) {
-		retlist = g_list_append(retlist, strdup((char *)vs.getText()));
-		vs++;
+	/* Use indexed access rather than popError()-driven iteration:
+	 * popError() on the ListKey stops the whole walk as soon as ANY
+	 * single element fails to resolve, silently dropping every
+	 * element after it, even when they are perfectly valid. Indexed
+	 * access lets us simply skip a bad element and keep going.
+	 *
+	 * But each INDIVIDUAL element also carries its own error state:
+	 * when a fragment of the input list doesn't parse as a valid
+	 * verse reference at all (e.g. a dictionary/Strong's-number key
+	 * or some other non-scriptural bookmark text accidentally fed
+	 * in here), Sword does not simply omit it -- it silently falls
+	 * back to some arbitrary boundary position (observed: the very
+	 * last verse of the Bible) and reports that as a "resolved"
+	 * element without necessarily failing the whole ListKey. Left
+	 * unchecked, that spurious element then looks like a genuine
+	 * match for whatever verse it landed on. Skip any element whose
+	 * own popError() is set, since it does not represent a real
+	 * reference from the input. */
+	for (int i = 0; i < count; i++) {
+		SWKey *elem = vs.getElement(i);
+		if (elem && !elem->popError())
+			retlist = g_list_append(retlist, strdup((char *)elem->getText()));
 	}
 	return retlist;
 }
--- a/src/gtk/bookmarks_menu.c
+++ b/src/gtk/bookmarks_menu.c
@@ -499,6 +499,31 @@ G_MODULE_EXPORT void on_crossref_popup_activate(GtkMenuItem *menuitem,
     xml_set_value("Xiphos", "misc", "crossref_popup",
                   settings.crossref_popup ? "1" : "0");
 }
+/******************************************************************************
+ * Name
+ *   on_tag_colorize_activate
+ *
+ * Synopsis
+ *   #include "gui/bookmarks_menu.h"
+ *
+ *   void on_tag_colorize_activate(GtkMenuItem *menuitem,
+ *                                 gpointer user_data)
+ *
+ * Description
+ *   toggle verse colorization by bookmark folder color
+ *
+ * Return value
+ *   void
+ */
+G_MODULE_EXPORT void on_tag_colorize_activate(GtkMenuItem *menuitem,
+                                              gpointer user_data)
+{
+    settings.tag_colorize =
+        gtk_check_menu_item_get_active(GTK_CHECK_MENU_ITEM(menuitem));
+    xml_set_value("Xiphos", "misc", "tag_colorize",
+                  settings.tag_colorize ? "1" : "0");
+    main_display_bible(NULL, settings.currentverse);
+}
 
 /******************************************************************************
  * Name
@@ -1263,6 +1288,10 @@ void gui_create_bookmark_menu(void)
 	gtk_check_menu_item_set_active(
 	GTK_CHECK_MENU_ITEM(menu.crossref_popup),
 	settings.crossref_popup);
+	menu.tag_colorize = UI_GET_ITEM(gxml, "tag_colorize");
+	gtk_check_menu_item_set_active(
+	GTK_CHECK_MENU_ITEM(menu.tag_colorize),
+	settings.tag_colorize);
 
 	gtk_widget_set_sensitive(menu.in_tab, FALSE);
 	gtk_widget_set_sensitive(menu.in_dialog, FALSE);
--- a/src/gtk/bookmarks_treeview.c
+++ b/src/gtk/bookmarks_treeview.c
@@ -1053,6 +1053,54 @@ void bookmark_debug_dump_colors(void)
 }
 
 
+/* Recursively search @folder (and its sub-folders) for a bookmark whose
+ * key resolves to @versekey_text. On match, returns "<immediate parent
+ * folder caption>: <bookmark caption>". Helper for
+ * bookmark_get_tag_info_for_key(). */
+static gchar *bookmark_find_in_folder(GtkTreeIter *folder, const gchar *versekey_text)
+{
+	gchar *folder_caption = NULL;
+	gtk_tree_model_get(GTK_TREE_MODEL(model), folder,
+			   COL_CAPTION, &folder_caption, -1);
+	GtkTreeIter child;
+	gchar *result = NULL;
+	if (gtk_tree_model_iter_children(GTK_TREE_MODEL(model), &child, folder)) {
+		do {
+			gchar *node_key = NULL, *node_caption = NULL;
+			gtk_tree_model_get(GTK_TREE_MODEL(model), &child,
+					   COL_KEY, &node_key,
+					   COL_CAPTION, &node_caption, -1);
+			if (node_key) {
+				/* leaf: a bookmark. Resolve its (possibly
+				 * multi-reference) key via Sword and compare
+				 * each resolved verse to versekey_text. */
+				GList *verses = main_parse_verse_list(
+					settings.MainWindowModule, node_key,
+					(char *)settings.currentverse);
+				for (GList *l = verses; l && !result; l = l->next) {
+					gchar *v = g_strstrip(g_strdup((const char *)l->data));
+					gchar *q = g_strstrip(g_strdup(versekey_text));
+					if (!g_ascii_strcasecmp(v, q))
+						result = g_strdup_printf("%s: %s",
+							folder_caption ? folder_caption : "",
+							node_caption ? node_caption : node_key);
+					g_free(v);
+					g_free(q);
+				}
+				for (GList *l = verses; l; l = l->next)
+					g_free(l->data);
+				g_list_free(verses);
+			} else {
+				/* sub-folder: recurse */
+				result = bookmark_find_in_folder(&child, versekey_text);
+			}
+			g_free(node_key);
+			g_free(node_caption);
+		} while (!result && gtk_tree_model_iter_next(GTK_TREE_MODEL(model), &child));
+	}
+	g_free(folder_caption);
+	return result;
+}
 gchar *bookmark_get_tag_color_for_key(const gchar *osiskey)
 {
 	GtkTreeIter folder, child;
@@ -1100,6 +1148,31 @@ gchar *bookmark_get_tag_color_for_key(const gchar *osiskey)
 		g_free(folder_color);
 	} while (gtk_tree_model_iter_next(GTK_TREE_MODEL(model), &folder));
 
+	return NULL;
+}
+/**
+ * bookmark_get_tag_info_for_key:
+ * @versekey_text: a Sword-formatted verse key string, e.g. "Rom 16:23"
+ *
+ * Finds the first bookmark, anywhere in the tree, whose (possibly
+ * multi-reference) key resolves to @versekey_text, and returns
+ * "<containing folder>: <bookmark name>" -- just the single
+ * immediate parent folder, not the full nested path.
+ * Returns NULL if no bookmark matches.  Caller must g_free() the result.
+ */
+gchar *bookmark_get_tag_info_for_key(const gchar *versekey_text)
+{
+	if (!versekey_text || !model)
+		return NULL;
+	GtkTreeIter folder;
+	if (!gtk_tree_model_get_iter_first(GTK_TREE_MODEL(model), &folder))
+		return NULL;
+	gchar *result = NULL;
+	do {
+		result = bookmark_find_in_folder(&folder, versekey_text);
+		if (result)
+			return result;
+	} while (gtk_tree_model_iter_next(GTK_TREE_MODEL(model), &folder));
 	return NULL;
 }
 
--- a/src/gui/bookmarks_menu.h
+++ b/src/gui/bookmarks_menu.h
@@ -44,6 +44,7 @@ struct _bookmark_menu
 	GtkWidget *remove;
 	GtkWidget *set_color;
 	GtkWidget *crossref_popup;
+	GtkWidget *tag_colorize;
 };
 typedef struct _bookmark_menu BOOKMARK_MENU;
 extern BOOKMARK_MENU menu;
@@ -64,6 +65,8 @@ void on_allow_reordering_activate(GtkMenuItem *menuitem,
 				  gpointer user_data);
 void on_crossref_popup_activate(GtkMenuItem *menuitem,
 				gpointer user_data);
+void on_tag_colorize_activate(GtkMenuItem *menuitem,
+			      gpointer user_data);
 void on_dialog_activate(GtkMenuItem *menuitem,
 			gpointer user_data);
 void on_edit_item_activate(GtkMenuItem *menuitem,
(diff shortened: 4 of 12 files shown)


karlkleinpaste:
First observation: It seems to work properly. That may seem like a trivial thing to say, but compared to the previous version a couple months ago, I'm pleased to be able to make the observation.

I have a couple further thoughts. No surprise there, I suppose.

1. I like the fact that, when user annotation highlight is enabled, it still shows a bookmark folder color 1-space block next to the verse, although the annotation highlight controls. Kudos for that. Turning off annotation highlight makes the whole verse highlight according to folder color.
Screenshot at 2026-07-03 17-33-23

2. However, the luminance test seems to be off. This verse is in a folder colored black (deliberately), and the resulting display is unreadable, when annotation highlight is turned off. I don't know if there's a way to improve this test, but I think you'd agree that "dark green text on black background" is far from a preferred colorization choice. It's probably significant, of course, that the verse is the current verse, which is why it's getting green in the first place. Perhaps being the current verse needs a special case consideration for luminance adjustment? I'm really not sure what to say about it other than that. I would appreciate your thoughts on it.
Screenshot at 2026-07-03 17-33-42

3. I have an _extensive_ set of bookmarks, after having been using Xiphos for 20 years; bookmarks.xml is ~250 lines long. So I've gone to the trouble to colorize my many folders, and unavoidably I've found myself re-using colors or using similar colors because, well, there are just so many colors out there. I have 2 folders that are somewhat different shades of blue, and another that's bright green. The result is a somewhat complicated-looking Gen 2.

Here's the thing: I can see that they're bookmarked because they're colored, but I have no reference for which bookmark in which folder.
Screenshot at 2026-07-03 17-34-00
Would there be a way to provide something like a mouse-hover (non-click) display in the status bar to show transiently (say) the bookmark name, or even folder+name? Not the whole folder-in-folder path -- my folders go 4 deep -- just the single containing folder. That would give me the clue to know where to go looking to find a bookmark I set 15 years ago and have long forgotten.

4. A final thing, but I must say it's important. I realize that colorization is explicitly driven by folders having color, and getting rid of color is a controlled by a folder being set for No Color. But there are times I want to share my screen with an audience (e.g. teaching Bible study) when I don't want everyone else in the room to have their eyes drawn to my particular colorizations. I think what's needed is an enhancement to turn on/off the entire bookmark colorization system, in Preferences -> General -> Options. Default on, because the colorized display is valuable, but able to be turned off so that no bookmark colorization happens at all. Then when I go to teach, I could just disable the whole thing, and others seeing my screen don't find their eyes scrambled by my unusual color choices.

The Options part of Prefs has gotten kinda long and dense, so I suspect part of this latter need is that the default (or should I say minimum) size of Prefs should grow by a dozen pixels, or maybe two.

If you like, I could merge this as it stands -- it's good and functional and valuable -- with the proviso to do some more for it.

karlkleinpaste:
Oh, no...
Well, I can't merge this as-is after all, because of a major problem in rendering. I should have looked for more examples before running into this.

image

The luminance test (I suppose) has failed drastically. The missing text is actually present, it's just invisible, as indicated when I mouse-sweep the entire chapter.

Screenshot at 2026-07-04 08-23-58

The text is present, but is white-on-white.

Also, this again shows that current verse highlight leads to unreadable text, as v.1 is dark green on medium blue. Current verse color highlighting may need to be avoided entirely when a bookmark folder color is selected. And this shows why a means for disabling the entire bookmark colorization facility needs to be available on demand.

karlkleinpaste:
This further exposes another problem:
- Only NET fails display this chapter correctly.
- Some bookmark color choices are not uniform across versions.

This image is ESV.

Screenshot at 2026-07-04 08-30-51

- There is no white-on-white text.
- There is only one bookmark-colored verse.

Indeed, only NET shows the errors.

My NET (not the standard module) is ThML, not OSIS, but by the time Xiphos is providing renderable text, it's all just XHTML, so the source origin shouldn't be what's affecting this.

karlkleinpaste:
I'm afraid this is getting worse.

I have just 2 bookmarks that include Rev 22.
```
grep Rev.*22 ~/.config/xiphos/bookmarks/bookmarks.xml
      <Bookmark modulename="" key="Deu.28.27; 1Sam.17.39; Isa.14.23; Rom.5.7; Rom.13.13; 2Tim.2.25; Matt.1.1; Psalm.12.7; Prov.18.24; Matt.5.47;17.20; 1Cor.13.8; Gal.5.12;6.4; Eph.2.1; Rev.22.19" moduledescription=" " description="KJV flaws"/>
      <Bookmark modulename="" key="Deut 4:2;12:32; Rev 22:18,19" moduledescription=" " description="Add to/take away"/>
```
But all versions are showing colorization as if I had a half dozen such bookmarks.

ESV.
Screenshot at 2026-07-04 09-03-48

BSB.
Screenshot at 2026-07-04 08-49-49

LAfricain:
I'll take a look at it next week, but I can already say that, strangely enough, I'm not getting the same errors as you—in Revelation, everything is displaying correctly.
Don't you have any other blue bookmarks?

karlkleinpaste:
Regrets that yesterday's "render whole books" PR has rendered this PR conflicted...probably deeply so. It will have to be reworked against the updated architecture where one display function for main window no longer does all the work; rather that function does just the initial HTML header generation and then farms out the rendering work to a separate RenderOneChapter function and another RenderWholeBook function which calls RenderOneChapter repeatedly.

It is possible that one type of error previously seen by me in this PR was caused in part by marginal module content involving paragraph breaks with improper trailing specifications. The paragraph ending problem has not one but rather two cases, both now handled. I hope I won't find any more in the future.

As I mentioned elsewhere, I am out of town for a few days but I'll take another look at all of these when I get back.

LAfricain:
Do you plan on working on this, or should I take a look? I have to admit, if you have a little time, I'd be happy to let you handle it, but let me know! I really like your commit that lets you scroll through the entire Bible.

I'm waiting for the last three PRs to be merged, and then I'll start translating the text that's still in English. Do you think it would make sense to release an new version once the three PRs are merged? These are some great new features!

LAfricain:
Finally I solved the conflict

karlkleinpaste:
I intended to test the current state of this. However, notwithstanding the claim by GH of "no conflics with base", ...
```
$ patch -p1 < ~/Desktop/1336.patch
patching file src/backend/sword_main.cc
Hunk #1 succeeded at 925 (offset 1 line).
patching file src/gtk/bookmarks_menu.c
Hunk #1 succeeded at 499 (offset 191 lines).
Hunk #2 succeeded at 1288 (offset 192 lines).
patching file src/gtk/bookmarks_treeview.c
patching file src/gui/bookmarks_menu.h
Hunk #1 succeeded at 44 (offset 1 line).
Hunk #2 succeeded at 65 (offset 1 line).
patching file src/gui/bookmarks_treeview.h
patching file src/main/display.cc
Hunk #2 succeeded at 97 (offset -2 lines).
Hunk #3 succeeded at 1549 (offset 123 lines).
Hunk #4 FAILED at 1671.
Hunk #5 FAILED at 1737.
Hunk #6 FAILED at 1748.
Hunk #7 FAILED at 1756.
4 out of 7 hunks FAILED -- saving rejects to file src/main/display.cc.rej
patching file src/main/settings.c
Hunk #1 succeeded at 1107 (offset 42 lines).
patching file src/main/settings.h
Hunk #1 succeeded at 106 (offset 2 lines).
patching file src/main/url.cc
patching file ui/xi-menus-popup.gtkbuilder
patching file ui/xi-menus.glade
Hunk #1 succeeded at 1793 (offset 9 lines).
patching file src/main/main.c
```
`src/main/display.cc` does not patch cleanly. The number of offset mentions disturbs me. As it stands, this isn't clean.

I simply don't trust GH itself. I trust `patch` because it's been around for 30 years without making mistakes.

You may need just a rebase.

LAfricain:
@karlkleinpaste done.

karlkleinpaste:
Regrets to say, this remains buggy: A great deal of colorization happens that should not be there.

I include screenshots from BSB Gen 1-3, plus stripped versions (to get rid of textual content) of my annotations from `settings.xml` and `bookmarks.xml`. For testing, you could replace your bookmarks temporarily with this set, and you could insert these `osisrefmarkedverses` into your `settings.xml`.





[stripped-annotations.xml](https://github.com/user-attachments/files/30663991/stripped-annotations.xml)

[stripped-bookmarks.xml](https://github.com/user-attachments/files/30663996/stripped-bookmarks.xml)

The most obvious demonstration of the problem is that I have neither annotations nor bookmarks in Gen 1. Yet the 1st image above shows 4 bookmark highlights.
```
fgrep Gen. stripped-*
stripped-annotations.xml:<markedverse label=" Gen.4.19" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.3.16" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.2.24" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.9.22" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.12.3" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.18.12" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.32.31" list="x"/>
stripped-annotations.xml:<markedverse label=" Gen.30.4" list="x"/>
```

karlkleinpaste:
Correction.
No annotations in Gen 1. But I have bookmarked Gen 1:5 (twice, in fact).
```
grep Gen\ 1: stripped-bookmarks.xml 
      <Bookmark modulename="" key="Gen 1:5; 8:13; Ex 40:2; Lev 23:24; Num 1:1,18; 29:1,33:38; Deu 1:3; Ezra 3:6; 7:9; 10:16; Neh 8:2; Esther 3:7; Ezek 26:1; 29:17; 31:1; 32:1; 45:18; Hag 1:1; 2:1; Mark 16:2; Luke 24:1; John 20:1,19; Acts 20:7" moduledescription=" " description="y"/>
      <Bookmark modulename="" key="Gen 1:5; 4:19; 8:13; Exod 12:18; 40:2; Lev 23:24; Num 1:1,18; 6:14; 29:1; 33:38; Deu 1:3; 24:4; 1Sam 1:2; 1Kings 6:25,27; 2Chr 3:11; Ezra 3:6; 7:9; 10:16,17; Neh 8:2; Esther 3:7; Ezek 20:26; 26:1; 29:17; 31:1; 32:1; 45:18; Hag 1:1; 2:1; Matt 28:1; Mark 16:2; Luke 14:18; 24:1; John 20:1,19; Acts 20:7; 1Cor 16:2; 1Tim 3:2,12; Tit 1:6; Rev 6:1; 9:12" moduledescription=" " description="y"/>
```
But not 1:1, 1:9, or 1:12, as colorized.

LAfricain:
Here is Claude's response:
It's a real bug, and I think I've figured out the cause. Take a closer look: Karl reports 4 colored verses in Genesis 1, even though he has no bookmarks or annotations in that chapter.

Here’s my hypothesis, which I think is pretty solid: `get_tag_color_for_versekey()` is called once per displayed verse (so ~31 times for Genesis 1), and with each call, it traverses the entire bookmark tree and calls `parse_verse_list()` for every bookmark it encounters. However, `parse_verse_list()` shifts the internal position of the same `VerseKey`/module object that the rendering loop is currently using to display the chapter—we had already identified this risk earlier (the “freeze” bug), and we had added a mechanism to save and restore the position around each call. But with dozens of repeated calls per chapter, on references that are sometimes malformed or complex, this constant back-and-forth on the same object can leave residual internal state in Sword (beyond the simple text being restored), with a risk that Sword’s “autonormalize” feature will reset an invalid position to… the very beginning of the Bible: Genesis 1. That would fit exactly with what Karl observes.

So there’s a real design issue that needs to be fixed, not just a minor bug: parse_verse_list() must never be called from within the verse-rendering loop. The correct solution: calculate a “verse reference → color” lookup table once, before rendering the chapter (or even the entire book in continuous mode), and then simply look up the entry in that table during rendering—without ever modifying the module’s position while displaying content.

What do you think?

LAfricain:
Found and fixed the root cause of the Genesis 1 false-positive
highlighting: bookmarks pointing at non-Bible modules (dictionary
entries, Josephus section numbers, etc.) were being fed to Sword's
verse-list parser as if they were verse references. Sword doesn't
fail cleanly on that input -- it silently reinterprets the bare
number/text as some arbitrary chapter:verse, which is what produced
the stray highlights. Bookmarks are now only resolved as verse
references when their module is empty (applies to all Bible modules,
as before) or is itself a Bible/commentary module.

Tested against your stripped bookmarks/annotations files -- Genesis 1
is now clean except for the two verses you do have legitimately
bookmarked there (1:1, 1:5), and annotations are displaying
correctly.

**Known remaining issue**: hyphenated verse ranges in a bookmark key
(e.g. "1 Cor 15:1-4") only highlight the first verse of the range,
rather than the whole range. The fully-qualified form ("1 Cor
15:3-1 Cor 15:4") works correctly. I'll follow up with a fix for the
shorthand form separately.

LAfricain:
Also fixed: hyphenated verse-range shorthand in a bookmark key (e.g.
"1 Cor 15:1-4") was only highlighting the first verse of the range,
since Sword's verse-list parser resolves that form to just its start
verse. The fully-qualified form ("1 Cor 15:3-1 Cor 15:4") already
worked and is untouched. Ranges now expand correctly regardless of
which form is used.

karlkleinpaste:
I am sorry, but this is still not successful. I have not changed either annotations or bookmarks, but BSB Gen 1 still colorizes too much for verses not being bookmarked at all. You should be able to see this in your own testing with my stripped bookmarks and annotations.




Gen.1.5 appears correct from `H->E` `#1C71D8`. But Gen.1.9 is the same color yet not bookmarked at all. Gen.1.1 is not bookmarked but has color from `H->C` `#35FF00`.

Gen.2.24 is correct from `PT` `#241F31` (with yellow user annotation overriding text). [ But it also appears under `H->C` `#35FF00` -- do you simply take the first (or last) match? ] Gen.2.18 is correct from `T` `#99C1F1`. But Gen.2.5 is colored from `H->C` `#35FF00` and should not be colored at all. Gen.2.21 is colored as though from `H->E` `#1C71D8` but is not bookmarked at all.

LAfricain:
Okay, I'll get back to it as soon as possible. Well, let's not get discouraged when we're so close to the finish line!

I don't see all of what you see. I do have your file with annotations and bookmarks; in fact, I had to remove the line breaks from the annotations so they would display correctly.


>Gen.1.5 looks correct starting at H->E #1C71D8. But Gen.1.9 is the same color even though it isn’t bookmarked at all.

On my end, it doesn’t appear in color (Gen 1.9).

>Gen.1.1 isn’t bookmarked but is colored starting at H->C #35FF00.

it’s colored on my end, it’s the first one in the H→C list.

>Gen.2.24 is correct based on PT #241F31 (with a yellow user annotation replacing the text). [But it also appears under H->C #35FF00—do you just take the first (or last) match?] Gen.2.18 is correct based on T #99C1F1. But Gen.2.5 is highlighted based on H→C #35FF00 and shouldn’t be.

Okay, I see that one highlighted on my end, too.

>Not at all. Gen.2.21 is highlighted as if it came from H→E #1C71D8 but isn’t added to the bookmarks at all.

I see that one, too.

karlkleinpaste:
> it’s colored on my end, it’s the first one in the H→C list.

Oh my goodness. You're right. I was grep'ing for "Gen " (G E N space) but Genesis is spelled out for that one. I hope I haven't represented many more false positives than are actually present. My apologies for that one. You may be closer than I realized.

LAfricain:
Correction gen 2.5 is present in H →C     <Bookmark modulename="" key="Genesis.1:1; Genesis.2:24; Jeremiah.3:6;Ezekiel.23:2;Leviticus.18:18;Matthew.25:1;2Samuel.12:8;Lev.18:20;lev20:10;Deu.22:22;Deuteronomy.25:5;Exodus.22:16;Deuteronomy.22:28;Exodus.20:17;Ruth.4:10;Proverbs.31:10;Deuteronomy.17:16;2Chronicles.24:1;2Chr.23:11;1Timothy.3:1;Revelation.6:1;Revelation.9:12;John.20:1;1Samuel.1:2;Genesis.4.19;1Corinthians.7:1;Galatians.3:28;**Genesis.2:5**;Matthew.19:5;Mark

LAfricain:
Sorry!
Gen 2.21 is not colorised... on my end... I think I did a mistake with 2.24.
Then Do you see something else?

LAfricain:
So if you notice anything else, let me know and I'll check it out, but as far as I'm concerned, everything's fine for now at my end.

karlkleinpaste:
I agree, we're getting quite close. I am finding some unusual module-specific issues.

For example, in my NET (in Xiphos repo), when in single chapter display, Gen 16 & 17 look normal.



When I change to whole book display, for some reason now it chooses to add colors to the first couple verses.



This isn't a "Module 1 good markup versus Module 2 bad markup" issue, it's all within one module. Thoughts or theories?

Well, ... wow. Again in my NET, Song gets to be a mess in whole book mode. Single chapter is fine.


Whole book goes crazy.


That blank space is actually text -- I can sweep it out with my mouse.


I just don't grasp how or why the text within a single module would induce so much wildly different/wrong choices.

karlkleinpaste:
Just noticed in the above... 4:14,15 (but not 16) and 5:1 have a grey background.

Even in single chapter mode, Song 5 is very confused.


How are different decisions about color being made, based on single chapter vs. whole book?

LAfricain:
One thing is clear: it's definitely a module issue! I just installed it and I'm seeing the same thing—but only with NET.

karlkleinpaste:
OK. I'll merge this and figure it out later -- probably this weekend -- as an adjustment on this. It's misbehaving in the same general way as the previous misadventures with paragraph endings, so it's probably just another instance of the same thing. I just have to find the particular oddity that NET expresses to trip the problem.

LAfricain:
Thank you! Now i'll work on the translation...

karlkleinpaste:
I have a suggestion for an enhancement on colorized bookmarked verses.

As I look at some chapter with several colorizations, I find myself wondering, "Which bookmark caused *that verse* to be colored"?

How hard would it be to add a marker, similar to user annotation `*u`, perhaps `*b`, that reacts to mouse-click by popping the bookmark that caused that colorization. Just the bookmark's label, that is, or maybe the folder name and label.

`<green>` ***b** [1] *u In the beginning, God created the heavens and the earth. `</green>`

Click *b, get the popup explanation.

You would need to get more back from the tag color detector than just the color, you'd need some text as well, to insert into the fake URL represented by what underlies *b.

Just a thought I had.

Maybe I'll take a shot at it myself.

LAfricain:
Very good idea!

PR #1340, opened 2026-06-30
fix: prevent </script> breakout in HtmlOutput anchor JS literal

hyder365:
`g_strescape()` escapes backslash, double-quote, and C0 control bytes but leaves `<`, `>`, and `/` untouched, so a sword:// fragment of `</script><script>PAYLOAD</script>` passes through unchanged when interpolated into the inline `<script>` block in `HtmlOutput()`. WebKit's HTML parser ends the `<script>` block at the literal `</script>` and starts a new `<script>PAYLOAD</script>` block, which executes arbitrary JavaScript in the `file://` origin that `wk_html_set_base_uri()` hardcodes.

The fragment can originate from any BibleSync UDP multicast peer on the LAN (`biblesync_glue.cc:179` constructs the sword:// URL from peer-supplied `bible` and `ref`), from `argv[1]`, or from a sword:// href click inside a malicious SWORD module.

Substitute `</` with `<\/` in the already-escaped anchor (OWASP recommendation) so the HTML parser can no longer see the end of the script block. `g_strescape` is kept in place to also handle the backslash-and-quote break-out case from #1337.

A self-contained reproducer is in `poc/reproducer.c` in this branch: it shows that `g_strescape("</script><script>alert(1)</script>")` returns the input unchanged.

Follow-up to #1337.

### DIFF
--- a/src/gtk/utilities.c
+++ b/src/gtk/utilities.c
@@ -1304,16 +1304,28 @@ HtmlOutput(char *text, GtkWidget *gtkText, MOD_FONT *mf, char *anchor)
 		// untrusted BibleSync navigation packets whose ref
 		// field carries the '#' fragment; without escaping a
 		// crafted value can break out of the string and inject
-		// script into the WebKit view.
+		// script into the WebKit view.  g_strescape escapes
+		// backslash, double-quote, and C0 control bytes but
+		// leaves '<', '>', and '/' untouched, so a value of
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
reproducer.c: you named them the same thing, and now i don't know if i should bother keeping the (small) one in this PR.
but i do appreciate the provision of an explicit means to reproduce.

hyder365:
Dropped the reproducer.

PR #1346, opened 2026-07-08
percent-encode clipboard selection in biblemap URL

acts-1631:
The "Open in Bible Map" action in the context menu builds a URL by
concatenating the clipboard selection into a fragment:

http://www.biblemap.org/#<selection>

The selection is not percent-encoded, so characters like spaces, #, &,
or newlines alter the URL structure passed to the system browser. A
second # in the selection hijacks the fragment portion.

The selection is now escaped with g_uri_escape_string() before being
inserted into the URL.


karlkleinpaste:
I deeply appreciate these sorts of analyses and repairs.
It's vaguely alarming how many such things can be discovered.
Thank you.

acts-1631:
Realistically I'm not sure how much of an attack surface a Bible study program is, but it's interesting and they're worth fixing.

Issue #1130, opened 2023-09-20
UX Enhancement Idea: Please Consider Hiding Language Selection Dropdown Menu  if Only 1 is Selected in Nesting Category 


profetik-777:
Assumption: 

The majority, perhas VAST majority are single lanuage users. 

If that is the case, my recommendation/request for this ux enhancement would be to bypass the language carrot drop-down and simply save the user from selecting the "ENGLISH" or whatever language they are using. 

![image](https://github.com/crosswire/xiphos/assets/112424307/3c839dc6-15e8-45a3-8acc-6a0a852b0b2e)

If they are multilingual, and end up adding a second language for that section - THEN the language drop-down becomes visible and is worth showing the user and comes back. 

Here is a demo of the idea/proposal with a video walk through. Thank you for the awesome software.

LAfricain:
It's a good idea. This already exists in Bibletime. But I think it would be best to offer the user a choice. Because the language is still useful when working with the source texts, in Greek and Hebrew or multiple translation. So in BT we have the choice, between category or languages, or both combined.

LAfricain:
@ariasrodolfo  it's what I said. My proposal is to add an option. But the main problem is nobody is still really working on xiphos...

LAfricain:
Can be closed.

PR #1349, opened 2026-07-20
modules: add configurable module-tree grouping (category/language/both)

LAfricain:
- new setting modules/grouping in settings.conf (0/1/2)
- main_load_module_tree_flat() and main_load_module_tree_by_language() in main/sidebar.cc, dispatched from main_load_module_tree()
- new combobox in Preferences > Modules Misc (prefs.glade + prefs.gtkbuilder)
- right-click on a category/language folder also offers the 3 modes

Addresses part of the display-customization ideas raised in #1130.

### DIFF
--- a/src/gtk/preferences_dialog.c
+++ b/src/gtk/preferences_dialog.c
@@ -81,6 +81,7 @@ struct _preferences_combo
 	GtkWidget *special_locale;
 	GtkWidget *font_prefs;
 	GtkWidget *display_columns;
+	GtkWidget *module_grouping;
 };
 
 typedef struct _preferences_color_pickers COLOR_PICKERS;
@@ -119,7 +120,7 @@ struct _preferences_check_buttons
 	GtkWidget *statusbar;
 	GtkWidget *alternation;
 	GtkWidget *justify_margins;
-
+	GtkWidget *show_hidden_modules;
 	GtkWidget *show_in_viewer;
 	GtkWidget *show_in_dictionary;
 	GtkWidget *show_devotion;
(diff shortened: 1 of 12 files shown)


LAfricain:
I realized too late that I created this branch from another branch that was still in progress! Two files were modified but aren't needed here sword_main.cc and sword_main.hh. I'm currently figuring out how to fix this.

I often run into this problem when creating a new branch that branches off from a branch other than the upstream branch.

LAfricain:
Sword repository is down. Windows build fails for this reason.

karlkleinpaste:
Being a paranoid soul, I tried re-triggering the failing Windows builds, just because it seemed a good idea.
[It didn't work](https://github.com/crosswire/xiphos/actions/runs/30207353002/job/90474553354), due to a dnf problem. I am ... confused.

LAfricain:
This looks unrelated to this PR — the failure happens in the Docker setup step (`dnf -y update --refresh && dnf install ...`), before any Xiphos source is even compiled. Our changes here only touch `.c`/`.cc`/`.h`/`.glade`/`.gtkbuilder` files.

Per `INSTALL.md`, the Windows cross-build requires Fedora 30 (since Fedora 31+ dropped webkitgtk). Fedora 30 is long EOL, and its repos have since been pulled from the standard mirrorlist and moved to `archives.fedoraproject.org`. A plain `dnf update --refresh` against an EOL release will fail once that happens — this is a common, recurring issue for Docker images pinned to old Fedora releases, not something caused by this patch.

This is almost certainly transient. Re-run the job first, in case it's a transient mirror hiccup.

LAfricain:
Now sword is down again. I wanted to suggest that we point to the sword mirror on GitHub to download the archive—it’s more stable. Lately, there have been recurring issues with the servers going down. That’s actually what worries me more broadly. I mentioned it on the mailing list, but I didn’t get any response.

karlkleinpaste:
That's very nice. A fine enhancement.

PR #1352, opened 2026-07-28
replace old hash-based anchor setting with modern scrollIntoView().

karlkleinpaste:
windows' webkit1 build has never had useful synchronization of current point of interest to widget top. this replacement for the old location.hash usage should make windows' webkit1 able to move to a named label as well as in later webkit.


karlkleinpaste:
failure. wk1 simply can't do it.

PR #1360, opened 2026-08-03
Test building Windows without GTK2

greg-hellings:
* Move to long-deprecated WEBKIT1 backend, but at least it gives us a
  shot at GTK3
* Use GTKTVEDITOR in the builds

TODO: Something is still pulling in gtk-2 in the images. That should be
sorted out before everything is fully merged

### DIFF
--- a/win32/CMakeLists.txt
+++ b/win32/CMakeLists.txt
@@ -95,6 +95,7 @@ if (WIN32)
     install(FILES
       ${SYS_ROOT_BIN}/libglade-2.0-0.dll
       ${SYS_ROOT_BIN}/libgailutil-18.dll
+      ${SYS_ROOT_BIN}/libgdk-win32-2.0-0.dll
       ${SYS_ROOT_BIN}/libgtk-win32-2.0-0.dll
       ${SYS_ROOT_BIN}/libjavascriptcoregtk-1.0-0.dll
       ${SYS_ROOT_BIN}/libwebkitgtk-1.0-0.dll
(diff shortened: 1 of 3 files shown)


LAfricain:
## Windows GTK3 build (`test/windows-build-gtk3`) — icons & Preferences dialog fix

### Context
After moving to GTK3/GtkBuilder/WebKit1 on this branch, two blocking issues showed up at runtime on Windows (tested via Wine): icons weren't rendering, and opening the Preferences dialog silently crashed the app.

### Diagnosis
The initial suspicion — leftover GTK2/libglade dependencies — was **ruled out**. Verified via `objdump`, source grep, and CI build logs: the build is fully GTK3/GtkBuilder, no libglade anywhere in the binary or the compile flags.

The actual problem: `win32/CMakeLists.txt` installs the GTK3 DLLs correctly, but was **missing three runtime resources GTK3 needs**:

1. The system icon theme (`hicolor`/`Adwaita`)
2. `gdk-pixbuf` loaders (needed to render SVG icons)
3. Compiled GSettings schemas (used internally by `GtkFileChooserButton`, which is used in the Preferences dialog)

Missing #3 caused a **fatal** error (`GLib-GIO-ERROR: No GSettings schemas are installed`) that killed the process the moment Preferences tried to open — hence the silent crash.

### Fix applied (commit on `test-nogtk2`, needs rebasing onto Greg's branch)
Added three `install()` blocks to `win32/CMakeLists.txt`, right after the existing SVG icon install section:

- `${SYS_ROOT_SHARE}/icons` → full icon theme
- `${SYS_ROOT_LIB}/gdk-pixbuf-2.0` → pixbuf loaders
- `${SYS_ROOT_SHARE}/glib-2.0/schemas` + compiled via `glib-compile-schemas` at install time → GSettings schemas

Verified in CI (GitHub Actions): build succeeds, all files confirmed present in the installed package.

### Remaining work (not done yet)
The resources are now bundled correctly, but GTK/GLib don't discover them automatically at runtime on Windows. Two environment variables need to be set for things to work:

- `GSETTINGS_SCHEMA_DIR` → path to `share/glib-2.0/schemas`
- `XDG_DATA_DIRS` → path to `share` (Windows-style path, `C:\...`)

Manually confirmed working (icons visible, Preferences dialog fully functional) once these are forced before launching `xiphos.exe`.

**Next step**: set both variables programmatically in `src/main/main.c`, before the `gtk_init()` call, computing the install path dynamically via `g_win32_get_package_installation_directory_of_module()` — rather than relying on GLib's automatic detection, which doesn't work reliably for this MinGW cross-compiled configuration.

greg-hellings:
Closing in favor of #1363 which includes this

PR #1363, opened 2026-08-05
Fix  Windows GTK3 build

LAfricain:
### Starting point
After moving to GTK3/GtkBuilder/WebKit1, two blocking issues showed up at runtime on Windows (tested via Wine): icons weren't rendering, and opening the Preferences dialog silently crashed the app. Several other issues surfaced and were fixed along the way.

---

### 1. Missing icons & Preferences crash

**Diagnosis:** The cause: `win32/CMakeLists.txt` installed the GTK3 DLLs but not three runtime resources GTK3 needs:
1. The system icon theme (hicolor/Adwaita)
2. `gdk-pixbuf` loaders
3. Compiled GSettings schemas (used internally by `GtkFileChooserButton` in Preferences)

Missing 3 caused a **fatal** `GLib-GIO-ERROR: No GSettings schemas are installed`, killing the process on Preferences open.

**Fix (`win32/CMakeLists.txt`):** added `install()` rules for the icon theme, pixbuf loaders, and compiled schemas.

**Follow-up bug:** even with resources bundled, GTK/GLib didn't discover them automatically at runtime — needed `GSETTINGS_SCHEMA_DIR` and `XDG_DATA_DIRS` set explicitly. Added this in `src/main/main.c`, computed dynamically via `xiphos_win32_get_subdir("share")`.

**Regression 1:** setting `XDG_DATA_DIRS` before the existing `g_get_system_data_dirs()` call (used for `SWORD_PATH`) redirected installed modules into the Xiphos install dir instead of the system location.

**Regression 2 (root cause, fixed properly):** `g_get_system_data_dirs()` **caches its result for the life of the process** on first call (documented GLib behavior). Calling it for `SWORD_PATH` — before *or* after setting `XDG_DATA_DIRS` — poisons that cache, and GTK's internal icon theme lookup (which uses the same cached call) never sees our override. Fixed by computing `SWORD_PATH` from `PROGRAMDATA`/`ALLUSERSPROFILE` directly instead of calling `g_get_system_data_dirs()` at all, so `XDG_DATA_DIRS`/`GSETTINGS_SCHEMA_DIR` can be set freely before `gtk_init()` without anything having poisoned the cache first.

**Result:** icons render correctly, Preferences opens without crashing, modules stay in the correct location. ✅

---

### 2. Editor ("studypad") crash — fatal `strdup`

**Symptom:** `xiphos va s'arrêter sur une erreur strdup` in `gtktextview_editor.c:469`, building `"<html><body>%s</body></html>"`.

**Diagnosis:** On glibc (Linux), `printf("%s", NULL)` prints `"(null)"` and continues; on MSVCRT/UCRT (Windows/MinGW), passing `NULL` to `%s` aborts the process immediately. Two call sites to `_load_text_into_buffer()` could pass `NULL` text:
- `_load_file()` — when reading a studypad file fails
- `editor_load_note()` — when a note/bookmark has no content yet (e.g. brand-new note)

A third call site already guarded with `text ? text : ""`; applied the same pattern to the two unguarded ones.

**Result:** editor/studypad no longer crashes. ✅

---

### 3. Daily devotional: enable GTK3 calendar on Windows

**Context:** a `#ifdef _WIN32` fallback showed a manual MM.DD text entry instead of the `GtkCalendar` popover, with the comment "Windows: GtkCalendar is broken." This was presumably true under the old GTK2/WebKit1 build.

**Fix:** removed the Windows-specific fallback in `src/gtk/dictlex.c` (widget creation `#ifdef _WIN32/#else/#endif`, and the `#ifndef _WIN32` guard around the calendar/button signal connects), so Windows now uses the same calendar+popover UI as Linux. The `entry_devotional` widget is kept unshown as internal date storage, matching prior Linux behavior.

**Result:** calendar popover opens and date selection works. ✅

**Known remaining cosmetic issue:** the devotional date button shows as a blank/gray box — no visible label text (not even the initial `"--"`) — though it's fully functional (clicking still opens the popover correctly, and the box size is fixed regardless of window resizing, ruling out a layout/reflow issue). This looks like a GTK3 "wimp" theme-engine text rendering quirk under Wine specifically, rather than an app bug — **pending confirmation on native Windows** (not yet tested outside Wine).

### DIFF
--- a/src/editor/gtktextview_editor.c
+++ b/src/editor/gtktextview_editor.c
@@ -523,7 +523,7 @@ _load_file(EDITOR *e, const gchar *filename)
 		g_error_free(error);
 	}
 
-	_load_text_into_buffer(e, text);
+	_load_text_into_buffer(e, text ? text : "");
 	if (text)
 		g_free(text);
 }
@@ -1383,7 +1383,7 @@ editor_load_note(EDITOR *e, const gchar *module_name, const gchar *key)
 	}
 
 	text = main_get_raw_text((gchar *)e->module, (gchar *)e->key);
-	_load_text_into_buffer(e, text);
+	_load_text_into_buffer(e, text ? text : "");
 	if (text)
 		g_free(text);
 
--- a/src/gtk/dictlex.c
+++ b/src/gtk/dictlex.c
@@ -621,16 +621,6 @@ GtkWidget *gui_create_devotional_pane(void)
 
 	widgets.entry_devotional = gtk_entry_new();
 	gtk_entry_set_max_length(GTK_ENTRY(widgets.entry_devotional), 5);
-#ifdef _WIN32
-	/* Windows: GtkCalendar is broken — show the MM.DD text entry instead */
-	gtk_entry_set_width_chars(GTK_ENTRY(widgets.entry_devotional), 6);
-	gtk_widget_set_tooltip_text(widgets.entry_devotional, _("Date MM.DD"));
-	gtk_widget_show(widgets.entry_devotional);
-	gtk_box_pack_start(GTK_BOX(hbox), widgets.entry_devotional, FALSE, FALSE, 0);
-	widgets.button_devotional_date = NULL;
-	widgets.popover_devotional     = NULL;
-	widgets.calendar_devotional    = NULL;
-#else
 	widgets.button_devotional_date = gtk_button_new_with_label("--");
 	gtk_widget_show(widgets.button_devotional_date);
 	gtk_widget_set_tooltip_text(widgets.button_devotional_date,
@@ -650,7 +640,6 @@ GtkWidget *gui_create_devotional_pane(void)
 	gtk_container_add(GTK_CONTAINER(widgets.popover_devotional),
                       widgets.calendar_devotional);
 #endif
-#endif /* _WIN32 */
 
 	/* prev button */
 	button_prev = gtk_button_new();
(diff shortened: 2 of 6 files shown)


greg-hellings:
@karlkleinpaste If you have no objections, I'll merge this. If we don't get major bugs related to it, we can consider beginning the removal of GTK2 specific code at that point as nothing remaining would use GTK2 pathways.

karlkleinpaste:
If you've made it work, please go ahead. I've just been watching this one from the sidelines. I would love to be rid of the gtk2 code someday. By itself, not having to maintain both gtkbuilder and glade files would be a blessing.

greg-hellings:
OK, it's merged in. I would think if we don't get any insurmountable bug reports after the next release we can then go and remove the GTK2 code.

Just in time to start thinking about moving to GTK4!! :eyes:

karlkleinpaste:
Do we care that the size of the GTK3-capable Windows installer grew by 15Mbytes?

Xiphos-windows-size

greg-hellings:
I wouldn't think that is a concern when we moved from the older to the newer GTK pack. I would expect a size bump. And in a world where the mobile apps I install on my phone routinely reach sizes larger than this and video games often reach many dozens of gigs of install I don't think our size is a concern.

However, this would be an opportune time to do an audit of the included files and DLLs to see if we're packing in any that are no longer necessary. We have sections in the installer for GTK2 vs GTK3 vs universally required support files. But it has been a long time since someone did an audit to check if all the ones in the universal list are actually required by both installs. The easiest way to do that is just to methodically remove each DLL from the `/bin` directory, launch the app, and check that it is still fully functional. Then repeat. Any that the app doesn't need can then be struck from the installer to ensure we're at minimum size.

LAfricain:
Yeah, I think that's normal—some themes were added to the package for displaying icons, and other stuff that I've already forgotten!

I've already started cleaning up the code a bit; I hope to finish on Monday.

PR #1365, opened 2026-08-07
adjust #1350.

karlkleinpaste:
the problem of paragraph endings <p/> appearing in text within <span> or <font> blocks is completely general; it induces premature closure of specifications, then psychotically re-introduces them later in the display (probable webkit bug, but not webkit's fault).  it doesn't happen just at the end of verses, it happens anywhere.  so the previous solution has been removed, replaced with a general search for <p/> when either highlight or bookmark colorization is active.  both "<p/>" and "<p />" (w/space) are handled because both are seen in existing modules.

also updated a couple nbsp; followed by a useless space, removing space.

### DIFF
--- a/src/main/display.cc
+++ b/src/main/display.cc
@@ -172,10 +172,6 @@ const char *bold_start = "<b>",
 	   *superscript_start = "<sup>",
 	   *superscript_end = "</sup>";
 
-gboolean paragraph_end_pending;
-gchar *para_type;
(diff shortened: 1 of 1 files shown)


karlkleinpaste:
I forgot I was using `<>` tags all through there. Oops. Hit the "..." on "adjust 1350" to get the proper text. Sorry, for anyone who cares.

PR #1368, opened 2026-08-08
provide for discovery of bookmark source of verse colorization

karlkleinpaste:
This adds to verse colorization per bookmark folder color by adding an element `*b` (similar `*u`) at the beginning of colorized verses whereby mouse hover induces a report in the previewer of all the bookmarks that reference this verse.

The tag map is modified so that the color string has been extended to include both tag color (background) as before, plus foreground color per luminance function, plus the name of the bookmark folder that references the verse. The content is carried as `#xxxxxx-#yyyyyy-LabelData` and then an arbitrary terminating string `@:@:@` is appended to delimit each element. All elements discovered will be appended to the list; the first pair is what is used for colorization.

When the user hovers over `*b`, the previewer reacts with a colorized list of bookmark names where the colorized verse was found.

- Significant changes to display.cc to extract the bookmark label from the tree, accommodate the updated tag color map, create and link the `*b` element.
- url.cc updated to provide a showBookmark capability, parallel to e.g. showUserNote.
- previewer.cc updated to recognize the new `*b` item and process the data string consisting of the content from the tag color map.

### DIFF
--- a/src/main/display.cc
+++ b/src/main/display.cc
@@ -1401,11 +1401,17 @@ static void build_tag_color_map(VerseKey *vk)
 	while (!g_queue_is_empty(stack)) {
 		GtkTreeIter *iter = (GtkTreeIter *)g_queue_pop_head(stack);
 		gchar *inherited = (gchar *)g_queue_pop_head(colors);
-		gchar *node_color = NULL, *node_key = NULL, *node_module = NULL;
+		gchar *node_color = NULL, *node_key = NULL, *node_module = NULL, *node_label = NULL;
 		gtk_tree_model_get(GTK_TREE_MODEL(model), iter,
-				   COL_COLOR,  &node_color,
-				   COL_KEY,    &node_key,
-				   COL_MODULE, &node_module, -1);
+				   COL_COLOR,       &node_color,
+				   COL_KEY,         &node_key,
+				   COL_MODULE,      &node_module,
+				   COL_DESCRIPTION, &node_label,
+				   -1);
+		gchar *escaped_label = (node_label
+					? g_uri_escape_string(node_label, NULL, TRUE)
+					: NULL);
+
 		const gchar *effective = (node_color && *node_color)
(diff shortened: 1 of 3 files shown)


LAfricain:
I just tested it, but unfortunately clicking on the bookmark doesn't open the Bible references in question—neither in chapter mode nor in full-book mode.
But otherwise, the idea is perfect, and I can clearly see the “b” that indicates the bookmarks when I manually navigate to the verse.

karlkleinpaste:
I was realizing this last evening. Bookmarks do not react to mouse click.

However, nothing I did in this PR touched bookmark _behavior_ at all. Not a thing. As you can see in the diff, 3 changes were made to display.cc (to extract the label; to change the colormap; to create the `*b` link), 1 to url.cc to react to showBookmark, and 1 to previewer.cc to react to type 'b'. Nothing in any bookmarks-reactive behavioral area.

I will experiment to see where the loss of capability occurred.

karlkleinpaste:
Well... I don't know how I could have done it. But `git reset --hard HEAD~` in my source tree shows that in fact this change caused the problem. Right now, I have no idea how. Looking.

karlkleinpaste:
Fixed. Mildly foolish mistake. I had created (or so I thought) a "showBookmark" capability for passagestudy URLs. But such a thing was already in use for ... showing bookmarks.

Replaced my "showBookmark" with "showBookmarkSource".

karlkleinpaste:
I need an opinion. I'm wondering if I've laid this out inconsistently, as a UI matter.

`<green>` `*b` [1] `*u` In the beginning God created the heavens and the earth. `</green>`
`<green>` [1] `*b` `*u` In the beginning God created the heavens and the earth. `</green>`

The first is what I created. Belatedly, I'm wondering if the second is more appropriate.

Thoughts? Should both `*u` and `*b` be after the verse marker?

LAfricain:
I'd say the second option—u and b next to each other?

PR #1373, opened 2026-08-13
Fix crash of Paralle view tab

LAfricain:
I think the double navigation bar is an XML file issue; I cleared it out, and now there's just one bar again. The fix for the crash seems pretty basic to me—I ran about twenty tests locally and in a VM, and I didn't experience any crashes; before, it would crash after just two tests. I hope that's it!


karlkleinpaste:
Crosswire is not answering (again). So Windows builds failed. No big deal.
I'll test the patch before merging.

PR #1381, opened 2026-08-18
codeclean: license headers, dead code, GResource migration

LAfricain:
Addresses part of #894. Three cleanup items from that issue are covered here;
two more intrusive ones (`xiphos_html` alias/macro rewrite, global/extern
variable decoupling) are intentionally left for separate follow-up PRs

## 1. Outdated FSF postal address in license headers (item 7)

Replaced the obsolete address block:

```c
 * You should have received a copy of the GNU General Public License
 * along with this program; if not, write to the Free Software
 * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA.
```

with the FSF-recommended pointer:

```c
 * You should have received a copy of the GNU General Public License
 * along with this program; if not, see <https://www.gnu.org/licenses/>.
```

- **196 files** updated: `.c` / `.h` / `.cc` / `.cpp` / `.hh` sources and
  `CMakeLists.txt` / `.cmake` build files (comment-only change, both `*` and
  `#` comment styles).
- `COPYING` (verbatim GPL-2.0 text) and obsolete/unused `.po` entries were
  deliberately left untouched.
- One file (`src/backend/gs_stringmgr.cpp`) had a truncated variant of the
  address and needed a separate targeted fix.

## 2. Commented-out / dead code (item 2)

Removed code left behind as comments across `src/`:

- **216 single-line** `// statement;` remnants (stale `g_free()` calls,
  leftover widget/`extern` declarations, etc.) — verified via a random
  30-line sample before bulk removal, no false positives found.
- **9 larger `/* ... */` blocks** — old implementations superseded by
  current code in the same file (e.g. a `glade_xml_get_widget(...)` block
  in `navbar_versekey_parallel.c`, replaced further down by the current
  `UI_GET_ITEM(...)` call), reviewed and removed individually rather than
  by pattern-matching.
- **56 files**, 401 lines removed. No `#if 0 ... #endif` blocks existed in
  the tree (already clean).

## 3. GResource for GtkBuilder UI files (item 4)

Scope for this pass: the 13 `.gtkbuilder` files plus `gtk_webedit.ui`
(icons/pixmaps left for a follow-up PR).

Previously, each dialog located its `.gtkbuilder` file on disk at runtime
via `gui_general_user_file()` and loaded it with
`gtk_builder_add_from_file()`. These are now compiled directly into the
`xiphos` binary via `glib-compile-resources` and loaded with
`gtk_builder_add_from_resource()`:

- New `ui/xiphos.gresource.xml` manifest, resource prefix `/org/xiphos/ui/`.
- New CMake build step in `src/gtk/CMakeLists.txt` generating
  `xiphos_resources.c`, linked into the `xiphos` target.
- **29 call sites across 19 files** switched from
  `gtk_builder_add_from_file` / `gtk_builder_add_objects_from_file` to the
  `*_from_resource` equivalents; now-unused `glade_file`/`gbuilder_file`
  path variables removed at each site.
- A dead `#if GTK_CHECK_VERSION(3, 10, 0)` fallback branch removed from
  `link_dialog.c` (no longer relevant once loading via resource).

**Follow-up fix (same PR):** several `.gtkbuilder` files reference images by
relative path (e.g. `<property name="pixbuf">logo.png</property>`).
`GtkBuilder` resolves that path relative to the resource's own bundle
directory once loaded via `gtk_builder_add_from_resource()`, rather than
relative to the file's location on disk as before — so those images had to
be added to the GResource bundle too, or the icon silently fails to load.
Affected: prefs, module manager, and splash-screen icons/logos (9 images
total, found via a headless GTK smoke test that loads every bundled
`.gtkbuilder` file and inspects the resulting `GtkImage` pixbufs). Two of
the nine images didn't yet exist under `ui/` and were copied over from
`pixmaps/` (`splash.png`, `gs2-48x48.png`), following the project's existing
convention of keeping real copies (not symlinks) under `ui/` for Windows
compatibility.

Not touched, deliberately out of scope:
- `src/editor/editor.c` has two functions (`editor_insert_sword_link`,
  `editor_insert_link`) referencing an undefined macro
  `LINK_DIALOG_UI_FILE`. They live in the `#else` branch of
  `#if defined(USE_WEBKIT_EDITOR) && defined(USE_WEBKIT2)` and are never
  compiled in the default (WebKit2) configuration — pre-existing latent
  dead code, unrelated to this migration, worth its own issue.
- Icons/pixmaps beyond the 9 pulled in above (general app pixmaps still
  load from disk) — candidate for a dedicated follow-up PR.

## Asset cleanup

Two apparently-unused images were identified:
- `ui/gs2-48x48-old.png` — no references anywhere in the tree (code, UI,
  CMake, desktop file, docs).
- `help/{C,fr,fa}/figures/gtkhtml3-vs-mozembed.png` — a screenshot from the
  old GtkHTML3-vs-Mozilla-embedding comparison; no active `.page` file
  references it (only a stale `TODO` note and obsolete/unused `.po`
  entries).

## Deferred

Left for separate, more focused PRs due to their larger surface area and
harder-to-verify behavioral risk:
- **Item 5** — replace the alias/macro layer in `src/xiphos_html` with the
  real underlying object.
- **Item 6** — eliminate global/`extern` variables and decouple exposed
  objects' internal implementation.


karlkleinpaste:
The latest windows builds have ballooned another 6Mbytes. In looking this over, I see that there is an entire subhierarchy `share/icons/Adwaita`. I'm not sure if this is brand new to the build or not, but it includes editor icons, weather icons, emojis, ... countless bits of inapplicable-to-xiphos, useless crud.
```
78304470 Aug 16 14:40 xiphos-4.4.0.75-win64.exe
84006681 Aug 20 16:12 xiphos-4.4.0.79-win64.exe
```
Is there a recent screw-up in CMakeLists.txt? If there's good reason to grow the build, sure, fine -- but it used to be in the 60M range, then gtk3 took it to 78M, now suddenly we're at 84M.

LAfricain:
Yeah, you're right—that's a lot! We do need the Adwaita theme fot gtk3 but we don't need all those icons.

PR #1382, opened 2026-08-20
Fix for #921: Caret in the text

### LINKED ISSUE #921: Caret in the text
Currently, it's impossible to copy a part of the text using keyboard only. It would be very helpful to show caret in the biblical text and enable selection with Shift+arrows.

LAfricain:
Enable WebKit caret browsing in Bible text and commentary panes (fixes #921)

Restricted to TEXT_TYPE and COMMENTARY_TYPE panes: enabling caret
browsing on every WkHtml instance simultaneously (dictionary,
sidebar, etc.) was found to interfere with keyboard caret navigation
(Down/Page_Down) in the active pane, even though the affected pane
kept receiving every key event normally.


LAfricain:
Oops, I just realized I synced the caret branch with master! I hope that doesn't cause any conflicts...

karlkleinpaste:
This change appears to have ruined `^F` as "find" in the text.

If you have just switched to a tab, then `^F` works. But if you are navigating within a book, `^F` stops being useful for "find."

This is not true for tabs that contain only a dictionary or only a genbook. This is only for the Bible pane.

karlkleinpaste:
This may be due to my use of gtk-key-theme set to "Emacs." Hence, any time the mouse has made the text pane active, Emacs-style keybindings are in effect, one of which is `^F` for "forward character." This explains why `^F` works for "find" when e.g. I have clicked a tab, such that the text pane of the Bible is not active with the keyboard.

PR #1388, opened 2026-08-22
Fix italics leak with per-fragment div-balancing

LAfricain:
Third attempt at the introMaterial() italics leak (#921 follow-up).

Previous attempts each fixed the leak but caused a new regression:
- Balancing the whole accumulated intro string and stacking missing </div> closes at the very end (original #1385) shifted block-level boundaries enough to break WebKit's column-break and anchor-scroll positioning in multi-column, whole-book rendering (Karl, #1385 comments).
- Wrapping every verse's text in a defensive <span style="font-style: normal;"> (the fix for that regression, #1387) stopped italics bleeding onto verse text, but is suspected of interfering with WebKit's last-line detection for text-align: justify, producing over-justified lines (Karl, #1385 comments, reproducible with Justify margins + 4 columns + whole-book rendering; not yet reproduced locally on this end).

This attempt balances each fetched OSIS fragment (chapter 0 and/or thisChapter "verse 0" text) individually, right where it is fetched, instead of accumulating corrections at the end of the whole intro block. The missing closing tags land immediately next to their own content rather than stacked far from their opening, which should be far less disruptive to surrounding block structure. The verse-text wrapping span is removed entirely -- no longer needed, and no new element is introduced around verse text at all.

Local testing: italics stay confined to the chapter intro on modules with rich introductions (e.g. FreCrampon), and Karl's column/anchor scenario (4 columns, whole-book rendering, Deu 24:17) checks out. Could not reproduce the justification issue locally to verify the fix for it; requesting Karl test this branch before merge.

### DIFF
--- a/src/main/display.cc
+++ b/src/main/display.cc
@@ -1212,26 +1212,46 @@ GTKChapDisp::introMaterial(SWModule &imodule, int thisChapter)
 				started_intro = true;
 			}
 
-			g_string_append(intro,
-					(settings.imageresize
-					 ? AnalyzeForImageSize(buf, CURRENT_COLUMNS,
-							       GDK_WINDOW(gtk_widget_get_window(gtkText)))
-					 : buf));
+			/* Issue #921: SWORD's OSIS filter can emit an opening
+			 * <div type="subSection" ...> in this fragment with no
+			 * matching closing </div> here (its eID apparently falls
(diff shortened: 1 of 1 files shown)


LAfricain:
Well, I just realized that the current version of Frecrampon has a really big problem—I don't understand how I missed it—so you can't use it as an example. I had an old version on my computer. I need to update it.

